PR #26, opened 2015-02-19
updates conda recipe version info

ddale:
This is a simple change to the conda recipe, making use of templating to dynamically set the conda package version information based on git tags. If you set `git_tag: v5.0.0`, for example, you will get a conda package called `pymca-5.0.0-np19py27_0`. As of this commit, if `git_tag: master`, it would produce `pymca-5.0.0-np19py27_10`, indicating the 10th commit after the v5.0.0 tag.

I recommend leaving `git_tag: master` committed in the repo, and making temporary local modifications in the working directory when creating conda packages for any particular release.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: pymca
-  version: master
+  version: {{ environ.get('GIT_DESCRIBE_TAG', '')[1:] }}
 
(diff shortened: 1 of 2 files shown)


vasole:
I have followed your recommendation, although it would be desirable to be able to specify the version/tag/commit from the command line without having to edit any file.

PR #30, opened 2015-09-30
MaskImageWidget and GLSupport

t20100:
- MaskImageWidget avoid reset zoom when drawing mask + opengl backend option (not yet paning enabled)
- GLSupport: Refactor + added internal features


vasole:
The changes to source under third-party have to be avoided. The idea is 
to have that software "as it is". If there are changes needed, they have 
to be reported to the upstream project.

Armando

t20100:
You're right.
This was just compile-time warning removal.

Thomas

PR #42, opened 2016-10-06
Adds ability to fit to numpy arrays

aaron-parsons:
Hi Armando,

We have a use case which requires us to operate on a numpy array instead of a file. The data is already loaded in memory. We will use the methods in this pull request to integrate PyMca with our data parallel, big data processing pipeline Savu: https://github.com/DiamondLightSource/Savu. Creating this new data "type" seemed the best way to achieve this; I'd appreciate your thoughts!

Best wishes,
Aaron

### DIFF
--- a/PyMca5/PyMcaIO/NumpyStack.py
+++ b/PyMca5/PyMcaIO/NumpyStack.py
@@ -1,0 +1,71 @@
+#/*##########################################################################
+#
+# The PyMca X-Ray Fluorescence Toolkit
+#
+# Copyright (c) 2004-2014 European Synchrotron Radiation Facility
+#
+# This file is part of the PyMca X-ray Fluorescence Toolkit developed at
+# the ESRF by the Software group.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


vasole:
Hi Aaron,

I am going to take a look.

Please note that the fastest way to deal with a stack of data is to use the FastXRFLinearFit module with a properly configured configuration file (about 40000 spectra/second).  In particular the fitMultipleSpectra method:

https://github.com/vasole/pymca/blob/master/PyMca5/PyMcaPhysics/xrf/FastXRFLinearFit.py

As you see you can call it with a numpy array.

Armando

aaron-parsons:
Hi Armando,

Wow! Thanks. That's pretty cool. I'll look into using this too. I hadn't realised this was an option.

The other method then would just be for people who collect high count rate data, of which we have a few.

Thanks for taking a look.

Best,

Aaron

#####################
Aaron Parsons
Data Analysis Scientist
I14: Hard X-ray Nanoprobe
Diamond Light Source
Harwell Science and Innovation Campus
Didcot
Oxfordshire
OX11 0DE

Phone:+44 (0) 123577 x8805
#####################
________________________________

vasole:
Since I have no copyright transfer declaration from your side, I tool the liberty to replace ESRF copyright by Diamond copyright into your file.

I made some other minor modifications. I hope not to have broken anything at your side.

aaron-parsons:
Thanks. I've checked it and it all works fine from my side.
________________________________

PR #45, opened 2016-11-21
[WIP] Use glReadPixels with RGB format when there is no alpha channel

t20100:
Add check to do glReadPixels with RGB format when there is not alpha channel in the buffer.
Quickly tested with and without alpha channel, but requires more testing.


vasole:
Should I integrate this PR? It is supposed not to break anything and in cases where it was broken (no alpha) it has a chance to get it working.

t20100:
As far as I remember, I had a case where the alpha channel was missing in the buffer.

vasole:
Object3D module deprecated.

PR #48, opened 2017-01-10
Use silx DataView widget

PiRK:
If silx is installed, use the `DataViewFrame` widget instead of `NumpyArrayTableWidget`, to show data when user clicks on "Show information" on a HDF5 dataset.

I leave the status as work in progress, because it is possible that the `DataView` module will be moved to another package than `silx.gui.widgets` before the coming release.


PiRK:
Now that DataViewer has a new view showing HDF5 metadata (shape, attributes...), we could remove the other two tabs in the "show info" widget. See https://github.com/silx-kit/silx/pull/553/commits/6a6248b8e5270b01ccf81c8c38783b432564d044

vasole:
- HDF5DatasetTable.py should not disappear. I cannot warrant somebody else is not using it.
- I have no means to disable using silx. To warrant a smooth transition, I intend to pass a --silx option to PyMca. Even if silx is disabled, the DataViewer would be still using silx and not the default table view.
- Perhaps this is for silx itself. Wen the dataset interpretation is spectrum, the default visualization should not be an image.

PR #49, opened 2017-01-20
[WIP] ScanWindow based on silx PlotWindow

PiRK:
This is an early PR to discuss about how to implement plugins for a silx PlotWindow. 
The solution I have come up with tries to duplicate as little code as possible. 

A plot proxy class handles the changes in the plot API  by overloading  a few methods (getActiveCurve, getAllCurves...) to provide the expected signature to the legacy plugins. All methods that have not changed between PyMca and silx are handled by the proxy's __getattr__ method.

All the plugin related code is implemented in the `PluginToolButton` class. This button must be initialized by passing the plot as a parameter. If the plot is a *silx* plot, the plot proxy class is used.

The `ScanWindow` class  only needs to inherit `silx.gui.plot.PlotWindow` and add the `PluginToolButton` to the toolbar (and all additional tools needed that are not implemented in PlotWindow).

At the moment, only legacy PyMca  plugins are supported. Do we expect that in the future new plugins will be coded using the silx plot API (`x, y, legend, info, params = getActiveCurve()`)? 
This would probably require an additional attribute in the plugin class, to specify which API is to be used.


vasole:
Initial discussion points:

- Instead of systematically merging info and plot dictionaries, one could add a keyword legacy with different behavior in case of being set to None, True or False.

- Effectively, the replot keyword is to be replaced by the resetzoom silx equivalent. I would write a universal function looking for replot in the keyword arguments and replacing it by resetzoom prior to the call.

PiRK:
This should PR be considered a proof of concept and closed.

PR #56, opened 2017-02-21
Adds a parameter so that we can turn off file saving

aaron-parsons:
Hi Armando,

I am running PyMca inside some MPI as part of savu, and the file saving of the fit images, and the creation of the IMAGES folder is causing locking issues for me. Instead I will just grab the images from the output dictionary without needing to save.

I added a parameter (nosave) to McaAdvancedFitBatch in order to run off the file saving. Is this ok? Or was there a better way I missed?

Aaron


vasole:
Hi Aaron,

Some comments about your pull request:

- I would change self.nosave to self._nosave no make clear is not something to be considered public. If later one you need access to it in a programmable way, we can use a setter and a getter (yes I'm old fashioned).

- Please get rid of the incompatible-with-python3 print statement (line 104).

Armando

aaron-parsons:
Hi Armando,

Thanks.
1/ Good plan. I am also a fan of getters and setters for larger projects.
2/Sorry I thought I removed them all...
 Changes made in the latest commit.

Aaron

PR #57, opened 2017-02-24
Fixes some bugs

aaron-parsons:
Hi Armando,
This PR does two things:
1/ fixes a bug I introduced in the last PR.

2/ Since January, our beamlines output nexus files with multiple entries. These entries are very different, however the PyMca batch fit assumes all the entries will be the same and tries to batch over all of these entries, which fails.

I added a try-catch around this statement so it carries on. It's been tested by users and they are happy. Incidentally, now we know that the batch fit can do this, we might write something on our side to exploit it in the future.

Also, since the users who noticed this bug, also want to process their data on windows, they asked me to ask when the next windows release will be? Would it be possible to do a patch release for windows that fixes this bug?

Aaron

### DIFF
--- a/PyMca5/PyMcaIO/HDF5Stack1D.py
+++ b/PyMca5/PyMcaIO/HDF5Stack1D.py
@@ -153,11 +153,12 @@ class HDF5Stack1D(DataObject.DataObject):
                 #expect same entry names in the files
                 #Unfortunately this does not work for SOLEIL
                 for entry in entryNames:
-                    path = "/"+entry + ySelection
+                    path = "/"+entry + ySelection 
                     dirname = posixpath.dirname(path)
                     base = posixpath.basename(path)
                     try:
-                        if base in tmpHdf[dirname].keys():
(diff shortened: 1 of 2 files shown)


vasole:
2/ Is tricky. Please notice it is considering the set of HDF5 files as a "Stack".

Stacks are expected to be set of files where the "mcas" in each file are the columns of your map and each file correspond to a row. If you have 20 files each of them with 5 "mcas" that should generate a map of 20 rows by 5 columns. In any case, it is expected that all files have the same structure.

Can you provide me one of your files? The way I had foreseen to deal with several non-equivalent entries in one file was to actually select the relevant entries.

aaron-parsons:
Hi,

I've sent you an email with some test data since github didn't like it in the comments. 
I see what you mean, and yes that could be tricky. However, for our case we don't want to fit anything in other entry. Just the data in one entry, and to skip the other.

I see you accepted the pull request, but let me know if you do want something changed.

Aaron

PR #58, opened 2017-04-27
External images plugin

PiRK:
ExternalImagesPlugin version based on a silx PlotWidget.

The external image is not reduced in size to fit the stack image size (the scales are simply recalculated).

TODO: 
- [ ] RGBWidget needs to be modified to accept any image size. Currently an error is raised when two images of different sizes are added.
- [x] Handle origin different from (0, 0)


vasole:
I think lines 194, 196 and 197 of SilxExternalImagesPlugin.py need to be changed.

PiRK:
196, 197 fixed. 194 seems to work, but i can replace `pop` with `del`, it is equivalent.

PiRK:
I plan to do ROIStackPlugin in the same PR. I believe there is not much left to do, I can reuse SilxExternalImageWindow and just add the filter and background features.

vasole:
Please, profit to use Roi instead of ROI in new classes and file names in order to be consistent with the CamelCase conventions.

PiRK:
Regarding sending the image to the RGBCorrelator, I implemented  a resize function in the ExternalImagesPlugin, just to get it to work without touching the correlator for the time being.

I found a new major problem with my approach of drawing the mask transparent image to ensure the correct mask size: the silx mask widget always works with the active image,   and so does the colormap toolbutton. So I cannot control the colormap without loosing the mask.

I need to find another way. Either adjust the colormap tool button and/or the mask in silx to be able to work with an image identified by its legend, or not use the silx colormap tool button.

Another issue with my approach is that the mask is drawn on an image with no data (all zeros) so the mask tools based on value thresholds do not work.

vasole:
@t20100 

What about having as active a transparent image of the proper size? (Only for the external images where the size does not match) Would that work?

PiRK:
Most discussed features are implemented. The widgets still lack Y axis orientation synchronization with the existing PyMca widgets, and I haven't started working on the scatters.

Regarding handling scatter data, in my proof of concepts I did a month ago, I chose to make two separate widgets for image data and scatter data. This made it a bit easier to set up the mask widgets, without needing to   have two mask widgets and two sets of setters and getters for the mask in a single widget. 

On the other hand, juggling between two separate mutually exclusive widgets in a GUI is also challenging.  So I'm not sure which way to go. I would recommend dealing with this question in a separate pull request, as the current PR does not affect the existing PyMca widgets while the scatter mask will probably require to have silx both in the main GUI and in the plugins.

vasole:
Since the underlying application does not support scatter yet, then just implement the functionality for images. If everything goes well, we'll be able to switch to using silx in the main GUI and in the plugins since the Silx Alternate ROI Window should be able to replace the widgets used in main window GUI. Just do not forget to remove the [WIP] flag when you feel ready.

vasole:
- The crop is not working properly. If I crop the image below by first zooming to the actual image region:

beforecrop

I get:

aftercrop

It looks as if the offset (~1.5) would not have been taken into account.

- It would be desirable to have the copy-to-clipboard silx action in the plots using silx widgets.

PiRK:
I just found: it did not take the origin into account and there was an unnecessary rounding of the image limits (expressed in plot coordinates). So basically the crop only worked when the origin was 0 and scale 1.

Now it should work.

vasole:
Looks good. Can you add the copy-to-clipboard action too?

PiRK:
Done.

vasole:
Bug in the Silx Alternate ROI Window. To reproduce:

- Open the alternate ROI window,
- Select a mask
- Move the slider to select one of the alternative images

PiRK:
I have a feeling that it is related to something else that I'm working on with Thomas: https://github.com/silx-kit/silx/pull/836

The funny thing is that if the mask is drawn on the PyMca side and as long as the mask widget is not opened in the silx widget, there is no bug.

PiRK:
The bug is fixed. It was caused by removing the image before changing it, when the slider is moved. This caused the mask to be deleted in the silx widget.  
Maybe the mask behavior needs to be corrected on the silx side.

vasole:
There is still a problem. When moving the slider in the Silx Alternative ROI, the existing mask, if any is lost.

PiRK:
I don't see this issue. But I noticed that the mask stays is not drawn in my widget until the mask icon is clicked for the first time.

vasole:
The problem happens if the ROI widget associated to the Silx window has not been shown.

It seems a problem with some silx optimization at some point. As a workaround, for the time being you can systematically show it docked.

PiRK:
I have a simple workaround for the error we saw when the mask widget send a maskUpdated signal with an empty mask at init. I just ignore empty mask (array of size 0) instead of making our widget emit the `sigSelectionMaskChanged` event.  

I also found the reason why the mask seems to be deleted when the slider is activated: the mask is not actually deleted, but the data image is redrawn on top of the mask, on the same layer z=1. This layer is hardcoded in the mask widget in silx. We could modify it to use `activeImage.getZLayer() + 1`.

When the widget is open, I guess the mask is redrawn after the image changes, so the problem is not visible.

My workaround is to use z=-1 for the background and z=0 for the data.

vasole:
Well, it seems to work, but I still do not catch how then the external images work.

In my view, both, the external image widget and the mask image widget should be able to accept a "background" image, an image and a mask and I would not like to test for the silx version in the immediate future.

I do not think we are forcing z layer to be positive. Probably working with -1 for the background, 0 for the active image and 1 (hardcoded) for the mask should work. Please, can you give a try at that? If unsuccessful I will accept your pull request as it is.

PiRK:
The external images are added as background images, underneath a tranparent stack data image. The mask stays therefore always on top of the visible image. 

I already tried what you suggest, and it works (https://github.com/vasole/pymca/pull/58/commits/10461786551b44475b8c8c022c8418b8fe29c07a)

vasole:
Merged. I would like to get rid of all those warnings when fabIO is not installed.

PR #66, opened 2017-06-14
Motor positions in stack

PiRK:
This PR implements a plugin to show motor positions in a stack.

- [x] proof of concept for image data
- [x] read motor positions in EDF
- HDF5 data files (implemented in #72)


vasole:
- In case of not having any positioners information, the plugin should just say it by opening an information dialog .
- It would be convenient to select some motors in the comboboxes instead of leaving them empty. That would help the user understand what is going on.

PiRK:
Most comments have been addressed. The only pending issue is that the motor positions widget is not automatically closed when the plot is closed. I'm investigating on solutions to do that (maybe make the MotorInfoWidget a QMainWindow or a QDialog instead of just a QWidget?)

PR #67, opened 2017-06-26
stackClosed

PiRK:
Implement a stackClosed() method for plugins, to be notified by stack widget.

Should I implement this method for most plugins, to close all widgets when the main window is closed?


vasole:
It would be convenient to have it in use ASAP. I am considering to have a default implementation in the lines of

``
if hasattr(self, "widget") and (self.widget is not None): self.widget.close()
``

PiRK:
Ready for a review.

PR #76, opened 2017-07-26
Make Hdf5Widget compatible with SPEC files

PiRK:
Make Hdf5Widget compatible with SPEC files if `silx` is available


PiRK:
It seems to be working.

PR #77, opened 2017-07-27
Improve positioners loading

PiRK:
This PR adds the option to read motor positions from a SPEC file, and makes `StackBase.setPositioners` more permissive regarding the data shape (as long as the data size is good).


PiRK:
Reading the mesh scan and loading the positioners from the same file seems to be working.

vasole:
Tested with data from BM23 (found under /tmp_14_days/sole/BM23).

The moved positioners found in microxas.scan are not loaded (third entry). Problems to be solved:

- Why are they not loaded?
- An information message dialog concerning the motors being ignored would be welcome.

PR #91, opened 2017-09-21
[Hdf5DatasetView] plugins support

PiRK:
This PR provides:
 - a `PluginsToolButton` to load plugins and serve as interface between the silx plot widget and `Plugin1DBase` (from PR #49 with minor modifications)
 - modification to `Plugin1DBase` to support silx plot widgets via  `PluginsToolButton`(form PR #49)
 - a `DataViewerFrame` modified to provide the plugin toolbutton, used in `QNexusWidget`

It makes minor changes to the right click menu behavior in `QNexusWidget` (open a menu also for groups and string datasets).
.

PR #96, opened 2017-10-11
[5.3] ScanWindow based on a silx plot

PiRK:
This PR implements a ScanWindow based on a silx PlotWindow. Most of the code comes from unmerged PR #49, but there are new updates due to recent deprecations in silx (PlotActions refactoring).

It also replaces the legacy PyMca `PlotWindow` with the silx `PlotWindow` in various plugins accessible from the `ScanWindow`.

PR #91 should be merged first (this branch is based on it, because it implements PluginsToolButton).

To do in separate pull requests:
 - stop using `LegacyScanWindow` (still used in `McaWindow` & `XMCDScanWindow`)
 - test and fix `XMCDWindow` (heavy use of `.dataObjectdict` for accessing curves)
 - add a plugin toolbutton to all new silx  `PlotWindow` instances

### DIFF
--- a/PyMca5/PyMcaCore/Plugin1DBase.py
+++ b/PyMca5/PyMcaCore/Plugin1DBase.py
@@ -176,7 +176,12 @@ class Plugin1DBase(object):
         """
 
         if silxPlot is not None:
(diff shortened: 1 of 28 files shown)


PiRK:
This can be merged.

PR #100, opened 2017-10-17
[5.3] PlotWindow

PiRK:
Use a silx PlotWindow instead of a PyMca PlotWindow.

Some PlotWindow directly related to ScanWindow have already been converted in #96.

TODO in separate PRs:

 - deal with `McaAdvancedFit` (more complicated than all the other widgets)
 - `ImageView`: use the silx one or keep this one? Not used in PyMca but in external plugins. One notable change: signature of `__init__`
 - add a `PrintPreview` toolbutton to all these `PlotWindow`s (currently the print icon is the *silx* one)
 - `MaskScatterWidget`: depends on `MaskImageWidget`. Should we try to use `SilxMaskImageWidget` (different API)

### DIFF
--- a/PyMca5/PyMcaGui/math/StripBackgroundWidget.py
+++ b/PyMca5/PyMcaGui/math/StripBackgroundWidget.py
@@ -255,6 +255,7 @@ class StripBackgroundWidget(qt.QWidget):
         self._x = x
         self._y = y
         self.update()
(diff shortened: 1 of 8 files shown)


PiRK:
I created issues for pending TODOs. So this PR can be merged.

PR #117, opened 2017-10-23
NXDataViewer with plugins

### LINKED ISSUE #99: [NXdataViewer] Missing plugins.
@PiRK

The dataview has the plugins but not the NXdataViewer

PiRK:
This PR adds a plugin tool button to the 1D viewer in NXDataViewer.

Closes #99

### DIFF
--- a/PyMca5/PyMcaGui/io/hdf5/Hdf5NodeView.py
+++ b/PyMca5/PyMcaGui/io/hdf5/Hdf5NodeView.py
@@ -45,6 +45,8 @@ from PyMca5.PyMcaGui.PluginsToolButton import PluginsToolButton
 from silx.gui.data.DataViewerFrame import DataViewerFrame
 from silx.gui.data import DataViews
 from silx.gui.plot import Plot1D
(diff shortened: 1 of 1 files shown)


vasole:
The plugins will need to be able to access "scatter data" because they are able to deal with unsorted data.

We'll need to implement that on a separated issue/PR.

PiRK:
This will require some modifications in `Plugin1DBase`, and to define how `getCurve`, `getAllCurves`... will look for scatter data that was not added as a curve. Or alternatively, rewrite `ArrayCurvePlot` and `NXdataScatterView` in PyMca so that scatter data is always added with `Plot1D.addCurve`.

vasole:
(I put @t20100 in copy)

I do not know what silx will decide. For PyMca, whenever there is only one axis, the NXdata should be represented as a curve and not as a scatter.

In the discussion we had concerning the need of a scatter object https://github.com/silx-kit/silx/issues/688, it was made clear to me that the need of having a scatter object was not for 1D visualization.

PR #121, opened 2017-10-24
Debian packaging à la silx

PiRK:
Adopt the silx package structure to re-use the silx build-deb.sh script. 

Work in progress.


PiRK:
I've made some progress, by uninstalling optional dependencies that where compiled with a newer version of numpy. After that, the tests pass for python2.7, but still fail for python2.7-dbg. I haven't found a way to disable the generation of dbg packages.

PiRK:
The failure while testing python-dbg was related to PySide. After uninstalling it, it works.

But I get a warning message "Cannot set sip API". I guess `python-sip`(`-dbg`) should be in the dependencies as well.

vasole:
The "cannot set sip API" is a warning that appears if PyQt4 has been imported prior to sip or if sip is not available. PyMca, unlike silx, supports API 1 and API 2.

PR #124, opened 2017-11-16
[5.3]MaskImageWidget plot API

PiRK:
Update the way MaskImageWidget interacts with the plots (RGBCorrelatorGraph and ProfileScanWindow)


PiRK:
This PR can be merged. 

It is a quick way of solving all the deprecation warnings by just adapting the legacy MaskImageWidget to the new plot API. This way, the appearance and the interaction (mask interaction using toolbuttons) of `QStackWidget` does not change.

I have not started looking at replacing `MaskImageWidget` with the `SilxMaskImageWidget` that is already used by some plugins. This will be a for another PR and will need more work, due to all the API changes.

vasole:
It sounds very reasonable. I'll test it and let you know.

PR #126, opened 2017-11-19
[console plugin] import from qtconsole for ipython>4

maurov:
I did a minor change for correctly importing from qtconsole if ipython>4
mauro

### DIFF
--- a/PyMca5/PyMcaGui/misc/QIPythonWidget.py
+++ b/PyMca5/PyMcaGui/misc/QIPythonWidget.py
@@ -71,8 +71,12 @@ else:
                 os.environ['QT_API'] = 'pyside'
 
 import IPython
-if IPython.__version__.startswith("2"):
+_ipy_main_ver = int(IPython.__version__.split('.')[0])
+
+if _ipy_main_ver == 2:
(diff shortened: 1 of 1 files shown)


vasole:
I do not see why the check > 4 is needed. It should be covered by the else with the added check abouot the possibility to import qtconsole.

```
if _ipy_main_ver == 2:
    QTCONSOLE = False
elif _ipy_main_ver > 4:
    QTCONSOLE = True
else:
    try:
        import qtconsole
        QTCONSOLE = True
    except ImportError:
        QTCONSOLE = False

if QTCONSOLE:
    try:
        from qtconsole.rich_ipython_widget import RichJupyterWidget as RichIPythonWidget
    except:
        from qtconsole.rich_ipython_widget import RichIPythonWidget
    from qtconsole.inprocess import QtInProcessKernelManager
else:
```

maurov:
@vasole you are absolutely right! the change introduced here makes no sense, sorry for that! I have traced back the problem I had on the beamline and this was simply due to a conflict between the conda environment and the PYTHONPATH variable, not the import from qtconsole. No need to merge, closing this.

PR #128, opened 2017-11-22
McaAdvancedFitBatch: add quiet option to __init__

tschoonj:
Hi Armando,

I have added a `quiet` keyword to McaAdvancedFitBatch to allow the caller to mute the discard output. When calling onNewFile with a data array one would otherwise get a lot of unnecessary output.

Best,

Tom


vasole:
I guess what you want is the opposite of what you have implemented (if quiet is true then do not log anything). Am I right?

tschoonj:
Yep that's right 😄 

I amended the commit just now...

tschoonj:
Thanks!

PR #131, opened 2017-11-29
[5.3]More bugs

PiRK:


PiRK:
I removed the WIP flag because I will not be working on it for a while. It can be merged whenever you want, but there is no hurry.

PR #135, opened 2017-12-07
Freeart integration

payno:


vasole:
The thing that bothers me is that the user needs to click on zoom reset to get the full picture. It is very confusing. For the rest, as far as there is no crash it is fine and we can react on user feedback.

For the stack plugin it is not what I meant (that can be achieved via the RGB correlator). I wanted to be able to process the whole stack. We can do that later on.

payno:
For the zoom issue we removed keeping the aspect ratio by default which seems to have an issue for this case when using matplotlib.

Issue #86, opened 2017-09-08
[ROI Imaging] Use new style NXdata


vasole:
When reading and writing PyMca still uses the ancient NXdata format.

The new format should be used for exporting data and both formats should be supported when reading.

vasole:
Closed by #136

PR #136, opened 2017-12-11
NXdata new style

PiRK:
Addresses #86

### DIFF
--- a/PyMca5/PyMcaGui/pymca/QHDF5StackWizard.py
+++ b/PyMca5/PyMcaGui/pymca/QHDF5StackWizard.py
@@ -36,6 +36,7 @@ from PyMca5.PyMcaGui.io.hdf5 import QNexusWidget
 from PyMca5.PyMcaCore import NexusDataSource
 from PyMca5 import PyMcaDirs
 
+
 class IntroductionPage(qt.QWizardPage):
     def __init__(self, parent):
         qt.QWizardPage.__init__(self, parent)
@@ -43,6 +44,7 @@ class IntroductionPage(qt.QWizardPage):
         text  = "This wizard will help you to select the "
         text += "appropriate dataset(s) belonging to your stack"
         self.setSubTitle(text)
+
 
 class FileListPage(qt.QWizardPage):
     def __init__(self, parent):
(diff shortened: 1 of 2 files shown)


vasole:
It seems this implementations gives up support of old style signals.

vasole:
I mean when reading.

PiRK:
Yes, you're right. My Mistake. The latest commit fixes this. If there is a signal attribute on the NXdata group, I assume that we are dealing with new-style NXdata. If not, I restored the old code for reading the old style signals.

PiRK:
I tested saving a "zoomed stack region as images" with PyMca 5.1.4, reinstalling PyMca from this branch, and reopening the file. It seems to work fine.

vasole:
The functionality is there and it works with both styles. Thanks!

The points that would be worth to be modified:

- The title is written as a bytes string
- The axes are written as an array of bytes
- One should check the encode attribute is there prior to perform the calls to it

vasole:
The last point refers to, for instance, the buffername parameter.

PiRK:
Regarding axes and title, it seems that actually ALL attributes and datasets are written as bytes.

Should I change all the occurences of `"...".encode("utf-8")` into `numpy.array("...", dtype=h5py.special_dtype(vlen=unicode)))`, or leave it as it is?

vasole:
If I am not mistaken, it should not be necessary to use the vlen trick for things not being a list. A single string should be directly writable. If if its a str written from Python 3 it will be utf-8 and if it is under Python 2 one has to make sure it is written as unicode (with utf-8).

PiRK:
Encoding a string or a unicode in python creates bytes.  H5py has no way of knowing that it was encoded as utf-8, especially if we restrict ourself to the  ASCII charset, so it will just follow its basic strategy of writing bytes as "Variable-length ASCII". 

For single string, we can avoid using the vlen trick by specifically using unicode: `u"..."` 
Then h5py will write them as "Variable-length UTF-8".

PiRK:
It seems that everything is written with the proper encoding now.

I noticed something subtle with h5py: depending on how attrs are written, it is sometimes not allowed to write unicode strings even when it is not a list (via the `attrs.create` API). 
This works:
`h5f["toto"].attrs["toto"] = u"toto"`
This causes an error:
`h5f["toto"].attrs.create("toto", u"toto")`

vasole:
If that problem can be reproduced in recent versions of h5py it should be reported to them.

PR #139, opened 2017-12-19
[WIP] "Fit all curves" plugin

PiRK:
I started with a simple prototype. Next we need to define which Fit manager to be used (I started with `Specfit` because it's the one I know best), and how to present results to users (one tab per curve presented as the usual table widget ? also a plot at the bottom of the tab ?).


vasole:
Some ideas that might help:

- The philosophy of other plugins is a menu with one action to configure and one action to apply.
- Output as an HDF5 with one curve per entry.
- Output as dictionary HDF5, JSON or ini with the dictionary dump of silx? (Probably just HDF5 as previous is safest)
- Possibility to generate HTML output as it is done with the McaAdvancedFit. Users are already asking me to be able to cut and paste the table with the results of the SimpleFit....

PR #146, opened 2018-02-02
Fit all curves

PiRK:
- [x] create a widget/GUI, similar to SimpleFitGui and StackSimpleFitAll
 - [x] implement saving fit results to HDF5 (draft function ready but it needs to be modified to save results for one fitted spectrum at a time)
 - [x] implement saving config as INI to HDF5
 - [x] create a Plugin to call the GUI and pass it the necessary data
 - [x] add bg to specfitfunctions

### DIFF
--- a/PyMca5/PyMcaGui/math/fitting/QScriptOption.py
+++ b/PyMca5/PyMcaGui/math/fitting/QScriptOption.py
@@ -243,7 +243,7 @@ class MyCheckField(CheckField.CheckField):
     def setdefaults(self, ddict):
         for key in self.dict.keys():
             if key in ddict:
-                if int(ddict[key]):
+                if ddict[key]:
                     self.CheckBox.setChecked(1)
                     self.dict[key]=1
                 else:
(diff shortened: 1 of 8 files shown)


PiRK:
All the basic components are in place, now we can start testing and improving. 

We still miss an output of the estimation (and therefore of the constraints).

vasole:
- The default functions indeed need a configuration widget to set the search parameters and default constraints.

vasole:
- I got a StringIO error under Python 3 when performing the fit. It is solved by #155.  
- The output needs quite a bit of formatting. We'll discuss it together.
- The copyright date should be 2018 for a new module created this year.

PiRK:
https://github.com/vasole/pymca/pull/146/commits/5ead67f8236d764941eb2af7a5ba665bf134db5b seems to fix my issue with the configuration being ignored.
But we need to check that this is a correct fix for all fit functions in PyMca, not only SpecfitFunctions.

vasole:
Without looking into details, I would say that if we return a dictionary when calling the configuration without arguments we should also send back a dictionary as input.

Issue #147, opened 2018-02-07
[fit] Read user fit functions on start up


vasole:
- [ ] Add compatibility layer with silx to be able to use same user defined functions
- [ ] Load fit functions found under fit directory at the same level as user plugins directory

vasole:
@PiRK 

Is this closed by the recent merge of the silx branch?

PiRK:
It is closed by PR #148 which was merged directly into the master.

PR #148, opened 2018-02-08
using silx fit theories in SimpleFitModule

PiRK:
Experimental attempt to address first goal in #147


PiRK:
This should be a single commit. I need to investigate why it includes all the silx plot branch.

PiRK:
Fixed.

PiRK:
TODO: provide a way for PyMca to know this is a silx function, due to difference in estimation signature.

PiRK:
Strangely, the error is originating in `Specfit.py`, which has not been touched in this PR. I'm still searching the cause.
The message is 

> ERROR on background and/or theory configuration



.

PiRK:
The error was caused by using a file with the same name as a PyMca standard module (SpecfitFunctions.py).

PR #153, opened 2018-02-16
Read the positions in analogous way to SPEC files.

### LINKED ISSUE #152: [nexus] read positioners when performing selections
Needed to provide the functionality associated to specfile motors.

vasole:
closes #152


vasole:
The SPEC header could only contain one value per motor corresponding to the positions at the beginning of the scan.

Should this functionality be limited to match the original functionality and to prevent problems?

vasole:
A solution could be to create a new positioners key and use the old ones the SPEC way.

PR #162, opened 2018-02-22
[DataViewer] cleaner way to replace views

PiRK:
This PR depends on silx PR  #1646


PiRK:
The required features seem to be merged in silx 0.7, so this is ready to be merged.

Issue #179, opened 2018-03-14
[XRF] Use HDF5 MCA description


vasole:
Automatically account for the information associated to an MCA written as NXdetector. This implies

- As it is done with other formats, the time, channels and calibration information has to be read.
- That information has to be accounted for in the MCA window.
- That information has to be accounted for in the batch fitting.
- That information has to be accounted for in the Fast XRF fitting plugin.

vasole:
closed by #182

PR #182, opened 2018-03-14
Use channels, calibration and time from HDF5 file

vasole:
Addresses #179 

- [x] The time, channels and calibration information is read.
- [x] Information accounted for in the MCA window.
- [x] Information accounted for in the batch fitting.
- [x] Information accounted for in the Fast XRF fitting plugin.


vasole:
This should be complemented by a unit test.

vasole:
Batch fitting correctly accounts for time.

vasole:
Everything working. Test to be split into tests with and without need of h5py.

PR #188, opened 2018-03-19
Stack Time info

### LINKED ISSUE #187: [ROI Imaging] Transfer time information
The time information should arrive to the MCA window if the stack contains it.

It should be able to provide the sum of the times or the average of the times.

vasole:
Closes #187


vasole:
Generated spectra already have associated time information.

Issue #166, opened 2018-03-02
Fit all plugin


vasole:
- Recover the labels from the curve information to add them to the NXdata plot
- Recover the used weights (if any) from the fit and add them to the NXdata plot

vasole:
- Allow to read the fit configuration back from the generated HDF5 file.

vasole:
#182 adds a readfp method to ConfigDict.py to allow to read the configuration from a StringIO,

vasole:
closed by #195

PR #194, opened 2018-03-29
[silx]Use sift from silx

### LINKED ISSUE #122: [silx] Use sift from silx
No reason to maintain two implementations.

PiRK:
Closes #122 

On my machine it works fine with PY3, and shows correct results on PY2 but segfaults on exiting the demo code in `SIFTAlignmentWindow`. 
The segfault is most likely unrelated to sift, I believe I messed up my system when I uninstalled a lot of packages to do debian packaging (It also segfaults on PY2 when I run the demo code in `McaWindow`).

I will do a few more tests to try to solve the issue on my machine, before removing the WIP tag.

### DIFF
--- a/PyMca5/PyMcaGui/math/SIFTAlignmentWindow.py
+++ b/PyMca5/PyMcaGui/math/SIFTAlignmentWindow.py
@@ -29,20 +29,20 @@ __contact__ = "sole@esrf.fr"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
 import os
-import sys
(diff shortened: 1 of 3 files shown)


PiRK:
I can see that there is more work, due to the change of signature of `LinearAlign(...)`

vasole:
If it is too difficult, I would give higher priority to #166

PiRK:
It was relatively easy. It just needs to be tested on Mac prior to merging.

PiRK:
I rebased this branch, to make the tests pass (it was missing travis and appveyor configuration files).

PR #198, opened 2018-04-06
 FitAll bugs

### LINKED ISSUE #197: [Fit all] Some bugs
- A direct configuration does not receive the active curve, therefore the SETUP of the strip does not work.
- The output file root is not editable. One can only choose output directory.
- After the fit is finished, the active curve is replaced by the last one.

PiRK:
Closes #197


PiRK:
All issues related to #197 are fixed. I can clean-up HDF5 file opening in dialogs in the same PR, after lunch.

PR #200, opened 2018-04-11
[silx]Replace DEBUG with logger

### LINKED ISSUE #181: [silx] Replace DEBUG by logger
Wait for 5.3.0 release to be out in order to update the silx branch and start.

PiRK:
Closes #181


PiRK:
I started working on Object3D, then I remembered that it might be deprecated soon. Is it OK to skip the modules that will be replaced by silx components (Object3D, PyMcaGraph)?

vasole:
Please, be carefull with the meaning of the DEBUG variable. Most of the times is logging, but in other cases the difference is raising an exception or not. See PluginsLoader.py case.

vasole:
Yes please, skip the Object3D module.

PiRK:
I fixed my mistake in PluginLoader. 

Every time there was a `if DEBUG: raise ...`, I replaced it with `if _logger.getEffectiveLevel() == logging.DEBUG: raise ...`.

Maybe this is not the best solution, because it mixes the concept of logging level and debugging mode. But the upside is that the behavior remains the same as before, in debugging mode we have both debugging messages and error raising.

To switch on debugging mode you can either do it in a single module (`_logger.setLevel(logging.DEBUG)`) or for the entire application (in `PyMca5/__init__.py`, set `_logging.basicConfig(level=logging.DEBUG)`).

We need to discuss this after you come back from holidays.

PR #201, opened 2018-04-12
Use silx icons when available

PiRK:


PiRK:
This works with PY3 / PyQt4, but does not work with Py2 / PyQt4 (the widget is shown blank, not icons and no labels).
On Henri's machine, it works with PY3 / PyQt5, and does not work  with PY2 / PyQt5 but for a different reason ("cannot build QPixmap without a QGuiApplication").

On both our machines, we can make it work by using `from silx.gui import qt` instead of `from PyMca5.PyMcaGui import PyMcaQt as qt`.

vasole:
If I follow https://doc.qt.io/qt-5.10/qtwidgets-widgets-icons-example.html it seems the usual size for the toolbar is 24x24. I was using 16x16 pixmaps from the time I needed compatibility with Qt 2.x ...

Issue #204, opened 2018-04-15
[debian] AttributeError: 'module' object has no attribute 'QGLWidget'


picca:
Hello, whne building pymca, I get this error

WARNING: autodoc: failed to import module u'PyMca5.Object3D.SceneGLWidget'; the following exception was raised:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/ext/autodoc/importer.py", line 140, in import_module
    __import__(modname)
  File "/<<BUILDDIR>>/pymca-5.3.0+dfsg/.pybuild/cpython2_2.7_pymca5/build/PyMca5/Object3D/SceneGLWidget.py", line 36, in <module>
    from . import Object3DRedBookFont
  File "/<<BUILDDIR>>/pymca-5.3.0+dfsg/.pybuild/cpython2_2.7_pymca5/build/PyMca5/Object3D/Object3DRedBookFont.py", line 153, in <module>
    class TestWidget(qt.QGLWidget):
AttributeError: 'module' object has no attribute 'QGLWidget'

on Debian unstable the Qt5 available is 5.10.
It seems that the pyqt5 removed this old compatibility layer...

vasole:
Strange. I am using Qt 5.10.1 under windows and the error is not there.

I will have to implement the workaround similar to the one in:

https://github.com/silx-kit/silx/blob/master/silx/gui/_glutils/OpenGLWidget.py

Two options, either you wait for the patch, either you totally remove the Object3D folder.

vasole:
@PiRK 

Do you have the time to implement the workaround in PyMca?

vasole:
It seems that replacing all occurrences of QGLWidget by QOpenGLWidget everything should work for your debian version but we need to implement the workaround for the case somebody insists on using PyMca with debian 8 or PyQt4.

PiRK:
It seems minimalistic patch would be:
```
diff --git a/PyMca5/Object3D/Object3DRedBookFont.py b/PyMca5/Object3D/Object3DRedBookFont.py
index b4b5125..bd20729 100644
--- a/PyMca5/Object3D/Object3DRedBookFont.py
+++ b/PyMca5/Object3D/Object3DRedBookFont.py
@@ -30,6 +30,13 @@
 import numpy
 from . import Object3DQt as qt
 
+if hasattr(qt, 'QOpenGLWidget'):  # PyQt>=5.4
+    _BaseOpenGLWidget = qt.QOpenGLWidget
[...]

PR #209, opened 2018-04-16
Use QOpenGLWidget with PyQt >= 5.4

PiRK:
Addresses  #204


vasole:
Sorry, I get errors with methods like setAutoBufferSwap and updateGL.

PR #221, opened 2018-04-25
Merge silx and old web information

vasole:
- Follow patterns similar to silx whenever applicable
- Reproduce the content of the old web site


vasole:
@PiRK 

- There are things from the old web site (still at http://pyma.sourceforge.net) that I do not manage to reproduce like having control over the Table of Contents on the left at the welcome page.

-  I get warnings about things not being included in any toc tree.

Any hint?

vasole:
OK. We merge this and open an issue for the rest.

Issue #207, opened 2018-04-16
[debian] please use setuptools requires in order to provide the build dependencies


picca:
Hello,

In order to build pymca and run the tests during the build, I needed to add all these dependencies.
It is possible to automatically generate the Debian binary dependency from the setup.py (like for pyFAI or silx). It would be great if you could provide this for pymca.

Then I could maintain a patch on top of this for Debian, in order to deal with the pyqt5 (dependency).
I am not sure thaht pyqt5 provide eggs infor for this purpose.

I got these informations looking at the autodoc sphinx part during the build. 

Cheers

```
                python-fisx-dbg <!nocheck>,
+               python-h5py <!nocheck> <!nodoc>,
+               python-h5py-dbg <!nocheck> <!nodoc>,
+               python-ipython,
+               python-matplotlib <!nodoc>,
+               python-matplotlib-dbg <!nodoc>,
                python-numpy,
                python-numpy-dbg,
+               python-opengl,
+               python-pyqt5 <!nodoc>,
+               python-pyqt5-dbg <!nodoc>,
+               python-pyqt5.qtopengl <!nodoc>,
+               python-pyqt5.qtopengl-dbg <!nodoc>,
+               python-pyqtgraph,
+               python-qtconsole,
                python-setuptools,
                python-sphinx,
                python-typing,
                python3-all-dbg,
                python3-all-dev,
                python3-fisx <!nocheck>,
                python3-fisx-dbg <!nocheck>,
+               python3-h5py <!nocheck> <!nodoc>,
+               python3-h5py-dbg <!nocheck> <!nodoc>,
+               python3-ipython,
+               python3-matplotlib <!nodoc>,
+               python3-matplotlib-dbg <!nodoc>,
                python3-numpy,
                python3-numpy-dbg,
+               python3-opengl,
+               python3-pyqt5 <!nodoc>,
+               python3-pyqt5-dbg <!nodoc>,
+               python3-pyqt5.qtopengl <!nodoc>,
+               python3-pyqt5.qtopengl-dbg <!nodoc>,
+               python3-pyqtgraph,
+               python3-qtconsole,
                python3-setuptools,
                python3-sphinx
```

vasole:
Wow, a lot of things there are either one or the other. For instance, qtconsole replaces ipython, pyqtgraph is certainly not needed.

Have you taken a look at our build scripts for Debian? They are supplied with the project:

https://github.com/vasole/pymca/blob/master/build-deb.sh

[...]

PR #231, opened 2018-04-28
Possibility to modify install requires.

vasole:
Related to #207

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -119,6 +119,27 @@ def use_cython():
 
     return True
 
+# check if GUI dependencies are to be considered install_requires
+def use_gui():
+    """
+    Check if GUI dependencies are requested.
+    """
+
+    if "WITH_GUI" in os.environ:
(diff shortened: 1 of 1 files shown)


picca:
Maybe it is better to avoid copy and just add the gui dependency to the initial minimal list of requires.

vasole:
I cannot  impose the GUI. There are people using PyMca just for analyzing XRF data in GUI-less workflows.

picca:
I was just telling that  it could avoid  copy of the dependency if instead of doing an if then else.
Just appending the gui dependency on user request :)

vasole:
I guess this is the simplest for you. Is it OK?

vasole:
I know. I just tried to make your life easier.

Please take a newer look at it.

vasole:
Compared to released v5.3.1, this pull request anticipates h5py as a hard dependency in subsequent releases. It changes nothing for debian.

vasole:
Waiting for the tests to finish prior to merge.

Issue #247, opened 2018-06-22
Deprecation warning when loading stack of EDF files


PiRK:
```
/users/knobel/.local/lib/python2.7/site-packages/PyMca5/PyMcaIO/EdfFile.py:649: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
  datatype)
```

It seems to be related to a recent numpy update: https://github.com/numpy/numpy/pull/9487 

> The rationale is that "string" means "unicode" in the brave new py3k world, and viewing unicode as binary is nonsensical.

Hundreds of warnings are printed.

vasole:
CClosed by #248

PR #248, opened 2018-06-22
Fix numpy.fromstring deprecation warning

PiRK:
In numpy 1.14, `fromstring` prints a deprecation warning when it is given binary data to process.

This PR uses `numpy.frombuffer` instead. Both functions have the same API in our use case, but `frombuffer` does not copy the data and returns a read-only array, so we need an additional copy.

The `numpy.frombuffer` function is available since at least numpy version 1.3.


PiRK:
Closes #247

vasole:
There are more appearances of ``numpy.fromstring``:

MaskImageWidget, RGBCorrelatorWidget, MarCCD, PilatusCBF, TiffIO, SilxExternalImagesStackPlugin

Can you add them to this PR?

PR #258, opened 2018-07-04
Settings dir

### LINKED ISSUE #257: Use ~/.pymca under Linux and MacOS
It is more standard to use that approach under those systems. 

The strategy should be:

- If ~/PyMca/plugins exists the program should keep using ~/PyMca
- If not, ~/.pymca should be created and used
- Windows should keep working as now as windows users are not used to access hidden files

PiRK:
Closes #257 

The code was duplicated in `PyMcaMain`. `FisxHelper` was already using the proper API.


vasole:
Thanks!

PR #268, opened 2018-07-11
[silx] Rebase onto master

PiRK:
Rebase silx branch onto master


PiRK:
There seems to be something wrong with this operation. WIP.

PiRK:
see #269

PR #269, opened 2018-07-11
[silx] merge master into silx

PiRK:
The rebase resulted in a lot of conflicts, and I don't really understand why. So I did a `git pull upstream/master` instead.


vasole:
SIlx branch merged into master via #274

PR #270, opened 2018-07-11
[silx] fix default backend deprecation warning

PiRK:


vasole:
Included in silx_merge.

PR #272, opened 2018-07-11
[silx]Don't set icon size in individual widgets

PiRK:
Icon size is already fixed in PyMcaQt.


vasole:
Already into silx_merge branch.

PR #286, opened 2018-07-13
Configure logging and fix deprecations

### LINKED ISSUE #275: Fix silx deprecation warnings
image

PiRK:
Closes #275 

This PR configures the logger for all applications and a few other modules.

I'm not sure if we should configure logging for all `main()` / `test()` functions in all modules in this PR. Maybe this can be done progressively, every time we edit a module for another reason?


vasole:
I would aim at having a long option command line --debug=INT to set the debugging level (at least for the main applications):

- 0 Not set
- 1 Debug
- 2 Info
- 3 Warning

PiRK:
I added the parameter. The way it is currently done, if both `logging` and `debug` are specified, the last one specified will probably over-rule the first one. This is the simplest way to implement it without adding too much code.

PR #293, opened 2018-07-17
[GUI] conserve plot lines and colors in McaAdvancedFit

### LINKED ISSUE #283: [GUI] McaAdvancedFit does not conserve plot lines and colors
This problem was also present in previous versions of PyMca.

To reproduce:

``python -m PyMca5.PyMca.McaAdvancedFit ./pymca/PyMca5/PyMcaData/Steel.spe ./pymca/PyMca5/PyMcaData/Steel.cfg``

- Show the legends
- Press Fit Again button
- Repeatedly click on the Peaks button.

The peak curves change color and linetype.

PiRK:
Closes  #283

All curves are replotted every time, so the easy way to fix this was to clear the plot before plotting again, which resets colors and style.


PiRK:
The alternative would be to add  `replace=True` in the first `addCurve`.

PR #301, opened 2018-08-21
[McaAdvancedFit] don't reset zoom when adding curves

### LINKED ISSUE #300: [Mca Advanced Fit] Switching peaks on or off should not reset the plot limits
PR #293 has the side effect of resetting the limits. They should be kept.

PiRK:
Closes #300. We need to explicitly set `resetzoom=False` in all `addCurve` calls.


vasole:
Very nice!

Issue #215, opened 2018-04-23
[2D Plugins] Define Plugin2DBase


vasole:
At this point we can add 1D and Stack plugins. We should be able to provide 2D plugins. The mechanism itself should be minimalist and the class implementing the methods should just hold a weakref to the plot. See Plugin1DBase.py as reference. Basically one needs:

- MENU_TEXT
- getPlugin2DInstance(plotWindow, **kw)
- the class with _plotWindow, getMethods, getMethodToolTip, getMethodPixmap, applyMethod

vasole:
closed by #297 and #302

PR #302, opened 2018-08-22
Provide signals to plugins

PiRK:
If a plugin has a `activeCurveChanged(prev, next)` or a  `activeImageChanged(prev, next)` method, they should be connected with the corresponding plot signal.


PiRK:
It seems to work. I will add documentation to `Plugin2DBase` in the other pull request (#215 )

Issue #280, opened 2018-07-11
[GUI] Take icons from PyMca_Icons


vasole:
I prefer to select the icons via PyMca_Icons in order to simplify the use of alternative icons.

I'm particularly concerned by the ROI icon and the print icon.

vasole:
closed by #304

PR #304, opened 2018-08-24
Use PyMcaIcons insead of silx icons for some icons

PiRK:
Possible solution for #280. The icon substitution can be done in a single function called by each plot window.


PiRK:
Ready to merge, unless you see another widget using the silx ROI&print icons.

Issue #346, opened 2018-10-10
[ROI Imaging] Provide Mask Scatter View Plugin


vasole:
The goal is to provide a way to deal with irregularly spaced stacks without having to rewrite the application.

Basically can work on a visualization based on point index and not any longer in dimension.

PiRK:
I found some two year old code that cold be relevant : https://github.com/PiRK/scripts/blob/master/MaskScatterWidget.py

I will see how it can be adapted to all new silx scatter features.

vasole:
Closed by #352

PR #352, opened 2018-10-16
Mask Scatter View plugin

PiRK:
Addresses #346 

The matplotlib backend seems to be close to its limits when displaying the ch09 stack as a scatter.

### DIFF
--- a/PyMca5/PyMcaPlugins/MaskScatterViewPlugin.py
+++ b/PyMca5/PyMcaPlugins/MaskScatterViewPlugin.py
@@ -124,9 +124,7 @@ class AxesPositionersSelector(qt.QWidget):
         self._initComboBoxes()
         i = 0
         for motorName, motorValues in positioners.items():
-            if numpy.isscalar(motorValues):
-                continue
-            elif self._nPoints is not None and self._nPoints != motorValues.size:
+            if not numpy.isscalar(motorValues) and self._nPoints is not None and self._nPoints != motorValues.size:
                 continue
             else:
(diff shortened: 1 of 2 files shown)


vasole:
Yes, I think OpenGL is almost mandatory.

PiRK:
It is almost ready, but still needs to be tested with real positioner data.

What is missing is to update the mask at the same time as the scatter when changing axes. It works if the mask tools widget is shown, but not if it is hidden (no active scatter handling if the widget is hidden).

PiRK:
It can probably be merged as it is. 

The few remaining problems with mask synchronization are hard to fix and are related to performance optimizations: the mask not updated on active scatter/image changes when the mask tool is hidden.
Whether it should synchronized all the time should be discussed on the silx side.

Anyway, the inconvenience is minor, as the mask is now properly drawn at widget initialization, and is properly redrawn whenever any synchronized widget makes any change to the mask. It only is a problem if this widget has the mask tools hidden and the user changes the axis, then the data points may move without the mask points following them.

PiRK:
I think I managed to find a workaround to force synchronization.

vasole:
A cherry on the cake it will be when we can specify background images with their positions and scales.

vasole:
Thanks!

PiRK:
The widget is currently minimalistic, but it could be improved to have a similar API to the `SilxMaskImageWidget`. 

The background image should be simple. But displaying other motor positions associated with each scatter point is not as trivial as with an image.

Issue #39, opened 2016-08-19
could you integrated the desktop file and the icon into pymca


picca:
Hello ARmando,

In the Debian packages, I have a bunch of .desktop files which are used to present pymca in the menu system.

could you add these files in the pymca package and install them during the build process under
{destdir}/usr/share/applications

and

{destdir}/usr/share/pixmaps

for the icons

override_dh_install:
    dh_install

```
# pymca
python setup.py install_scripts -d debian/pymca/usr/bin

dh_install -p pymca debian/edfviewer.desktop usr/share/applications
dh_install -p pymca debian/elementsinfo.desktop usr/share/applications
dh_install -p pymca debian/peakidentifier.desktop usr/share/applications
dh_install -p pymca debian/pymca.desktop usr/share/applications
dh_install -p pymca debian/pymca.xpm usr/share/pixmaps
dh_install -p pymca debian/pymcaroitool.desktop usr/share/applications
dh_install -p pymca usr/share/man
```

vasole:
Hi Fred,

Are you talking about my unofficial debian packages or about the official ones? My packages conflict with the official fisx package and they can only be considered of use if no official pymca package is available. 

You will be happy to read that we are working on a generic launcher. That will allow to have a single script.

Armando

picca:
Hello Armando, I speak about the official pacakge even if it miss a few keywords :))

> You will be happy to read that we are working on a generic launcher. That will allow to have a single script.

usually the best is to use the setuptools entry point. No ?


Fred

picca:
just for info pymca 5.1.2 entered debian unstable today :))

vasole:
The idea is that you can have something in the lines of:

pymca [options]
pymca viewer [options]
pymca batch [options]
pymca correlator [options]
...

with a single entry point. Something similar to what git it is doing.

vasole:
Closed by #121

vasole:
It seems it does not follow current Deian rules (no xpm files)

vasole:
Hopefully closed by #370

PR #370, opened 2018-11-13
Desktop icons

PiRK:
addresses #39 

I think this should work, but I'm not 100% sure (one size PNG file? relative icon name instead of absoute path?). 

From what I understood from https://wiki.debian.org/AppStream/Guidelines#Icon_best_practices, it is enough to provide the largest possible icon (256x256) and let the system handle the downscaling.

> icon of **at least** 64x64px in size ... This avoids accidential upscaling of icons.

silx has a SVG icon, so it does not have this PNG scaling issue.

@picca: is there is simple way to test that it is really working? If I build a debian package and install it, where should I see this icon appear?


PiRK:
It seems to be working on my machine 
pymcaicon

vasole:
That looks great!

PR #374, opened 2018-11-15
debian packaging for 5.4.1

PiRK:
It works for debian 8. We have not managed to do the debian 9 packaging yet.


PiRK:
I fixed a first issue, but I still have a problem with running the tests using the build (data dir not found). If I don't manage to progress on this by tomorrow morning, I will skip these tests. 

Anyway, it makes more sense to run the tests after completing the build and installing the package.

PR #396, opened 2018-12-21
[packaging] Add input file

### LINKED ISSUE #395: [desktop] desktop-mime-but-no-exec-code
Hello it seems that the Exec command line miss a %F in order to give the file  to the application.

vasole:
closes #395


picca:
No :)), but maybe there is a way to explain this :)

picca:
https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables

vasole:
OK. I think we are missing a .desktop file for the pymcapostbatch and/or rgbcorrelator.

They are the same script. I guess we should create just one of them.

vasole:
See #397

Issue #378, opened 2018-12-10
[QIPythonWidget]Importing ConsolePlugin  fails due to PyQt4.QtCore and PyQt5.QtCore modules both wrap the QObject class


PiRK:
When running `pymca --debug=1` on rnice, I get this error :
```
<type 'exceptions.IOError'> Problem importing module ConsolePlugin
<type 'exceptions.RuntimeError'>
the PyQt4.QtCore and PyQt5.QtCore modules both wrap the QObject class
<traceback object at 0x7faa16db73f8>
   File "/usr/bin/pymca", line 9, in <module>
    execfile(fname)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/pymca/PyMcaMain.py", line 1786, in <module>
    PyMcaMainWidgetInstance = PyMcaMain(**keywords)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/pymca/PyMcaMain.py", line 326, in __init__
    self.mcaWindow = McaWindow.McaWindow(backend=backend)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/pymca/McaWindow.py", line 93, in __init__
    info=info)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/pymca/ScanWindow.py", line 312, in __init__
    specfit, info, save)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/pymca/ScanWindow.py", line 166, in __init__
    directoryList=pluginDir)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGui/PluginsToolButton.py", line 129, in getPlugins
    PluginLoader.getPlugins(self, method, directoryList, exceptions)
  File "/usr/lib/python2.7/dist-packages/PyMca5/PyMcaGraph/PluginLoader.py", line 152, in getPlugins
    raise IOError(exceptionMessage)

```

I believe that the code  importing Qt binding in `QIPythonWidget` has not been updated recently, and might not be aligned with the order set in PyMcaQt. And we should test for `'PyQt5.QtCore' in sys.modules` rather than `"PyQt5" in sys.modules`.

vasole:
I guess what we can do is to default to PyQt4 under python2 ...

vasole:
closed by #402 and #403

PR #403, opened 2019-01-04
[GUI] Arrange priorities between QtConsole and IPython

vasole:


vasole:
Consider #378 as closed.

Issue #180, opened 2018-03-14
[HDF5] Provide output in HDF5 format


vasole:
It has to follow the NXprocess approach of the fit all plugin. That means

- Fits and maps saved as NXdata making use of auxiliary_signals
- Configuration used saved to the HDF5 file
- The configuration from the HDF5 file must be readable back and used
- The saved maps must be readable back wherever PyMca can currently read a list of EDF files.

PiRK:
Should we add a save button to all fit widgets ? Or only some specific fits ?

vasole:
What I had in mind was mainly XRF related stuff. It is true that it can also be made for the ROI Imaging simple fit plugin.

vasole:
Closed by #440

PR #415, opened 2019-01-18
H5fastfit

woutdenolf:
Part of issue  #180: HDF5 output for fast fitting
- Nexus utils for saving + unit tests
- Optional output formats: edf, csv, tiff, h5
- Optional diagnostics: data, model, residuals (not in GUI)
- Output naming conventions (compatible with old convention):
    For HDF5: dir/root.h5::/entry/process
    For others: dir/root/entry.edf
    Defaults (same as before):
         root = 'IMAGES'
         entry = 'images'
fastfitdialog
- Add bootstrap for development
- Share saving routine between GUI and script
- Refactor core Fast Fitting routine for readability

### DIFF
--- a/PyMca5/PyMcaGui/physics/xrf/FastXRFLinearFitWindow.py
+++ b/PyMca5/PyMcaGui/physics/xrf/FastXRFLinearFitWindow.py
@@ -264,9 +264,9 @@ class FastXRFLinearFitWindow(qt.QWidget):
         output['h5'] = int(self._h5Box.isChecked())
         output['overwrite'] = int(self._overwriteBox.isChecked())
         diagnostics = int(self._diagnosticsBox.isChecked())
-        output['save_data'] = diagnostics
-        output['save_fit'] = diagnostics
-        output['save_residuals'] = diagnostics
+        output['saveData'] = diagnostics
+        output['saveFit'] = diagnostics
+        output['saveResiduals'] = diagnostics
         return ddict
 
 class FastXRFLinearFitDialog(qt.QDialog):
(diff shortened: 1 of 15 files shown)


vasole:
The fast fit tries to be economic in memory usage and it is expected to works even with HDF5 datasets not fitting into memory. When generating the fit and residuals matrices, you seem to assume that twice the size of the actual map is going to fit into memory when even the map can be still lying on disk.

I would recommend to check if the memory can be allocated and, if not, proceed as if having selected HDF5 output in order to map the output arrays and fill them.

woutdenolf:
I cannot reproduce the Travis errors of this PR:
 https://travis-ci.org/woutdenolf/pymca/builds/483922089
or locally with the same version of python (3.6.3) and numpy (1.16.0)

vasole:
Line 271 in StackInfoTest. You are modifying the variable "values" over which you are iterating. At least in the past some versions of python complained about it. I think it would be better to use different names.

vasole:
McaStackViewTest and NexusUtilsTest are missing in tests/__init__.py

woutdenolf:
Changes:
-  Moved MCA stack access and output to their own modules (McaStackView and FastXRFLinearFitOutput resp.)
-  Handle MCA stacks of any dimension (work is done in McaStackView)
-  Precision: calculations are done in numpy.float and results are saved in numpy.float32 (unless the spectra itemsize > 4, in which case numpy.float64 is used)
-  Add more tests
-  Adapt coding style (e.g. camel case)
-  Small changes to the GUI:
fastfitdialog

vasole:
There are minor things to be changed, but they can go into separate issues.

Great stuff. Thanks a lot.

PR #440, opened 2019-03-29
Batchfit hdf5

woutdenolf:
The main purpose of this PR is to homogenize batch fitting and fast fitting in terms of output.
Most interesting additions are .h5 output and saving of diagnostics (e.g. fit residuals).
Apologies for the extensive refactoring.

Comparing GUI for batch and fast fit that provides identical output (in terms of structure):

slowfit

fastfit

Changes from a user perspective:
- optional output formats like for fast-fitting (most importantly .h5)
  .csv and .tif are disabled for multiprocessing because merging hasn't been implemented for those
- .edf output results in a single multi-image (like in fast-fitting)
  `OutputBuffer` has the option to save single-image edf's but this is not exposed by the GUI
- concentrations.txt file is only saved when fit files are saved

Some comments on implementation:
- Both batch fitting and fast fitting now share the same `OutputBuffer` (class that takes care of memory allocation and saving)
- `McaAdvancedFitBatch._processList` seems to support multiple non-HDF5 stacks but judging from `McaAdvancedFitBatch.__processStack` this cannot work I think
- The fit configuration is saved in .h5. However only one configuration can be saved while `McaAdvancedFitBatch` supports multiple configurations.
- PyMcaBatch has been refactored (e.g. to support bootstrap). This has only been tested on win32 and debian (non-frozen). Binaries on Mac seem to pass only roifit options, seems strange.
- `McaAdvancedFitBatch` has an `_obsolete` parameter to preserve the way data was saved before (testing purposes, currently disabled). Keep it or remove?

The files `McaAdvancedFitBatch` and `PyMcaBacth` contain several `TODO`'s which indicate some doubts I have on existing functionality (not introduced by this PR).


vasole:
This one is going to take quite a bit of time to verify ... I guess the safest, besides your foreseen obsolete parameter, will be to prepare LegacyPyMcaBatch and LegacyMcaAdvancedFitBatch using the versions prior to this pull requests to compare (results, speed and frozen behavior). We will be able to produce unit tests for both of them (except when frozen).

One thing not actually ever completed was to make McaAdvancedFit able to run as standalone using the same options as PyMcaBatch -or to make PyMcaBatch run without Qt- but that is another history.

I guess you are correct about multiple stacks probably not working. I am not sure about ever having tried it.

It seems you are making some assumptions that do not necessarily hold:

- about pieces of code not being used. For instance, the option not to save anything is used by Diamond that integrates McaAdvancedFitBatch in workflows going only through memory.

- units on output. I guess this can only be safely used if there are units on input.

- all the images in a single EDF risks not to be appreciated by many users. Probably user choices should be provided.

- All what PyMca can do with EDFs it can also do it using the TiffIO module.

woutdenolf:
We will discuss in detail, but some preliminary replies:
- Legacy versions: The `obsolete` flag was meant for just that, but I will create separate classes and add unittests to cross-validate.
- `nosave` flag: ok my bad, I though this was just for testing purposes. I assume they get the results from `savedImages`? Shall I add a `savedImages` property to mimic the old API (because memory is now managed by `outbuffer`)? Where can I find their code?
- "about pieces of code not being used": anything else than the `nosave` flag? If so it was unintentional.
- units: ok I will remove it from this PR but missing units (and uncertainties) are unacceptable from a scientific pov. I would even add `dimensionless` explicitly to the mass fractions for example. But lets keep that for other PR's.
- multi-page edf: fast fitting produced multi-page EDF's and batch fitting single-page. So I choose multi-page as the default for both. Shall I make single-page the default and add an option to the GUI? We are starting to get lots of options though...
- "All what PyMca can do with EDFs it can also do it using the TiffIO module with TIFF files": you mean with regards to multi page? The tiff part was just copied from the old fast fitting code.

vasole:
I can help with writing the unit tests. I can produce them for the current master.

My problem with the units is that to use "counts" when the input could be normalized data makes no sense. We can only afford to use work with units if we are sure about the units we get on input. I would leave it for the future. For instance, we should be able to make use of the (now clear) NeXus convention of appending _errors to the name of the datasets to properly deal with units and uncertainties.

Concerning Diamond use, what I know comes from PR https://github.com/vasole/pymca/pull/56. They said they take the results from the dictionary.

Yes, I meant that TiffIO supports multipage like EDF.

vasole:
Concerning your commit 698bc12 [batchfit] do not launch more subprocesses than files, the test has to be on number of spectra, not on files.

We have to be able to use multiple processors to process an HDF5 stack in a single HDF5 file. Other formats are not so important, but at the very least that one is a clear case we have to support.

vasole:
Hint: Use the mca step instead of the file step.

woutdenolf:
Addressed last comments:
- Units: removed
- Multi-page: use single page by default, add GUI option and allow multi-page for tif
- nosave: moved to the OutputBuffer (because of hdf5 memory allocation)
- concentrations.txt: put back (saving cannot be disabled from GUI when fitting with concentrations)
- overwrite: now optional in both GUI's
- Multi-process merging: supports all file formats
- LegacyMcaAdvancedFitBatch: old implementation, tested together with the new in StackInfoTest
- LegacyPyMcaBatch: old implementation (test are still missing StackInfoTest)
- XRFBatchFitOutputTest: separate tests for OutputBuffer (testing formats and memory allocation, nosave and overwrite)

This are the default GUI options:
fastfit
slowfit

woutdenolf:
I resolved the multi processing issue. I also added tests for PyMcaBatch to compare:
* McaAdvancedFitBatch
* LegacyMcaAdvancedFitBatch
* PyMcaBatch (single vs. multi processing, currently still coupled to the GUI)
* LegacyPyMcaBatch (not done yet)

The tests currently run on HDF5 and EDF input data (generated from Steel.spe).
I didn't manage to generate spec data that McaAdvancedFitBatch is supposed to be capable of handling. If you have an example of such a file, please send it.

So to be done: test spec data input + old GUI (LegacyPyMcaBatch )

woutdenolf:
Currently these tests are applied to an EDF and HDF5 stacks generated from Steel.spe (PyMcaBatchTest):
- FastXRFLinearFit vs. LegacyFastXRFLinearFit
- McaAdvancedFitBatch vs. LegacyMcaAdvancedFitBatch
- PyMcaBatch vs. LegacyPyMcaBatch (+ thread vs. single process vs. multi processing for new and legacy code)
What needs to be done:
- run the same tests on spec data (except for the fast fitting). I would need an example file to be able to generate the correct spec file for testing.
- enable tests on Mac (currently testing this in VM).

vasole:
Minor things.

- PyMcaBatch. The launching widget should hide and the progress widgets should be shown during the calculation.
- PyMcaBatch. The .dat files give different column headers compared to past (Al_K instead of Al-K)
- PyMcaBatch. Not exactly the same output values (differences of 0.001 %).
- PyMcaBatch. HDF5 Labels. I find weird to see spaces in the HDF5 file. I think the NeXus  recommendation is to use underscores and to avoid spaces. I have to make my mind if I want underscore everywhere or hyphen everywhere. Probably we could have underscores everywhere except for the .dat file for backwards compatibility.

vasole:
Confirmed http://download.nexusformat.org/doc/html/datarules.html

woutdenolf:
> * PyMcaBatch. The launching widget should hide and the progress widgets should be shown during the calculation.

It doesn't? How did you run it (OS, single/multi process, frozen or not)?

> * PyMcaBatch. Not exactly the same output values (differences of 0.001 %).

Yes, rtol in the tests 1e-5. I will check but I think this is because the `OutputBuffer` manages data allocation (float32 by default).

> * PyMcaBatch. The .dat files give different column headers compared to past (Al_K instead of Al-K)
> * PyMcaBatch. HDF5 Labels. I find weird to see spaces in the HDF5 file. I think the NeXus  

As long as the labels are consistent over .dat, .csv, .tif and .edf I ok with it (this wasn't the case before). My idea was to keep them as they are in HDF5 and remove spaces + decorate with w(...) and s(...) for the other formats.

vasole:
Running under windows. Command pymcabatch, not frozen.

vasole:
Do not worry about the numeric difference. The tolerance is OK.

woutdenolf:
> * PyMcaBatch. The launching widget should hide and the progress widgets should be shown during the calculation.

Hmmm. As you can see [here](https://github.com/woutdenolf/pymca/blob/79a317b435122770865c8de4c7e933f94a8028c0/PyMca5/PyMcaGui/pymca/PyMcaBatch.py#L1298) I hide the main widget with `self.hide()` except when running a single independent process. Not sure which is is not hidden...

vasole:
I think you hide the widget, then you start the process/processes and you ask it to show again the widget without verifying everything it if has finished.

woutdenolf:
> I think you hide the widget, then you start the process/processes and you ask it to show again the widget without verifying everything it if has finished.

Yes of course, I totally missed that. Fixed.

vasole:
OK. That one is fixed for multiple processes.

When having a single process, the widget is not hidden and one is allowed to start again leading to a crash.

woutdenolf:
Also the result is not shown in multi processing.

vasole:
For single process, after pressing OK for the notification dialog the widget has to hide and the progress bar widget has to be risen to make sure it is not hidden.

vasole:
- LegacyPyMcaBatch does not show the results at the end. It used to show the edfviewer but not the pymcapostbatch.

woutdenolf:
> * LegacyPyMcaBatch does not show the results at the end. It used to show the edfviewer but not the pymcapostbatch.

It seems to open the edfviewer shortly and than close (or crash).

woutdenolf:
> When having a single process, the widget is not hidden and one is allowed to start again leading to a crash.

It is actually the subprocess that crashes when it tries to handle an exception (caused by pyqt exception handling, I guess it tries to open a dialog to show the exception). This also happens on Linux btw:


```python
<type 'exceptions.IOError'> Unable to create file (unable to open file: name = '/data/id21/inhouse/wout/sota/pymcatestdata/output/xrfmap_xia00_0001_0000_0000_to_0004.h5', errno = 17, error message = 'File exists', flags = 15, o_flags = c2)   File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaGui/pymca/PyMcaBatch.py", line 1520, in run
    self.processList()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py", line 254, in processList
    with self.outbuffer.saveContext():
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaPhysics/xrf/XRFBatchFitOutput.py", line 729, in saveContext
    with self.bufferContext(update=update):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaPhysics/xrf/XRFBatchFitOutput.py", line 641, in bufferContext
    with self._h5Context(cleanup_funcs, update=update):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaPhysics/xrf/XRFBatchFitOutput.py", line 667, in _h5Context
    with NexusUtils.nxRoot(fileName, mode='a') as f:
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-2.7/PyMca5/PyMcaIO/NexusUtils.py", line 418, in nxRoot
    with h5py.File(path, mode=mode, **kwargs) as h5file:
  File "/mntdirect/_data_id21_inhouse/wout/dev/virtualenvs/rnice8/pymca/py27/local/lib/python2.7/site-packages/h5py/_hl/files.py", line 394, in __init__
    swmr=swmr)
  File "/mntdirect/_data_id21_inhouse/wout/dev/virtualenvs/rnice8/pymca/py27/local/lib/python2.7/site-packages/h5py/_hl/files.py", line 184, in make_fid
    fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 105, in h5py.h5f.create

QObject: Cannot create children for a parent that is in a different thread.
(Parent is QApplication(0x3505f20), parent's thread is QThread(0x1ac83d0), current thread is McaBatch(0x3901380)
Segmentation fault
```

For now I disabled single independent subprocesses on Windows to mimic the old behavior (hides window but application in blocked). The subprocess crashing is not as bad as the main GUI being blocked imo. So I would keep non-blocking and try to solve the issue of the crashing subprocess. Of course we could also use the multiprocessing polling strategy for single processes as well...

A comment on subprocesses: I now use `subprocess.Popen` for everything:

- multiprocessing: keeps the subprocess handles in a list and polls the list until done (same as legacy code)
- independent process: does not keep the subprocess handle so no way to know when finish (so don't hide window either). In the legacy code this was `os.system("... &")`. In fact this also exists on windows: `os.system('START /B ...')`. Anyway, `subprocess.Popen` replaces all this.
- blocking process: calls `wait()` after launch the process, thereby blocking the application (hide before and show after)

vasole:
> Of course we could also use the multiprocessing polling strategy for single processes as well...

I guess it would be the best since it might reduce the amount of possibilities to maintain.

The only case where it cannot be used is the case of a frozen MacOS application.

woutdenolf:
I revisited the launching of threads/processes for fitting and viewing. Tested manually for everything except frozen binaries on mac

Upside:
- more uniform
- not blocking the main process on windows

Downside:
- it takes a few seconds before the viewer pops-up for single process fitting (because single process is not treated like multiprocess).

These are the changes ("polling" means the main process polls the child processes, merges and/or shows results when they are finished, like was done for multiprocessing in the legacy code):

### Single process

|          |      Old      |  New  |
|----------|-------------|------|
| linux script | `os.system("..&")` | `suprocess.Popen()` + polling |
| linux frozen | `os.system("..&")` | `suprocess.Popen()` + polling |
| win script | `subprocess.call` | `suprocess.Popen()` + polling |
| win frozen | `subprocess.call` | `suprocess.Popen()` + polling |
| mac script | new thread | `suprocess.Popen()` + polling |
| mac frozen | new thread | new thread |

### Multiple processes

|          |      Old      |  New  |
|----------|-------------|------|
| linux script | `suprocess.Popen()` + polling | `suprocess.Popen()` + polling |
| linux frozen | `suprocess.Popen()` + polling | `suprocess.Popen()` + polling |
| win script | `suprocess.Popen()` + polling | `suprocess.Popen()` + polling |
| win frozen | `suprocess.Popen()` + polling | `suprocess.Popen()` + polling |
| mac script | `suprocess.Popen()` + polling | `suprocess.Popen()` + polling |
| mac frozen | exception | new thread |


### McaBatchWindow.plotImages: EdfFileSimpleViewer

|          |      Old      |  New  |
|----------|-------------|------|
| linux script | `os.system("..&")` | `suprocess.Popen()` |
| linux frozen | `os.system("..&")` | `suprocess.Popen()` |
| win script | same thread | `suprocess.Popen()` |
| win frozen | same thread | `suprocess.Popen()` |
| mac script | same thread | `suprocess.Popen()` |
| mac frozen | same thread | same thread |

### McaBatchGUI._showProcessResults: EdfFileSimpleViewer (after multiprocessing)

Always in the same thread

### McaBatchGUI._showProcessResults: PyMcaPostBatch (after multiprocessing)

|          |      Old      |  New  |
|----------|-------------|------|
| linux script | `os.system("..&")` | `suprocess.Popen()` |
| linux frozen | `os.system("..&")` | `suprocess.Popen()` |
| win script | `suprocess.Popen()` | `suprocess.Popen()` |
| win frozen | `suprocess.Popen()` | `suprocess.Popen()` |
| mac script | `os.system("..&")` | `suprocess.Popen()` |
| mac frozen (*) | `os.system("..&")` | `suprocess.Popen()` |

(*) never reached

woutdenolf:
As for the unittests, I have the multiprocessing tests for SPEC and HDF5 input disabled on MAC. In CI I occasionally get segfaults (not been able to reproduce locally).

vasole:
While for HDF5 it might be an issue, for SPEC the concurrent access should not be a problem.

In the past I was disabling multiprocessing when there was a single HDF5 file...

PR #449, opened 2019-04-20
OmnicMap IO: remove windows check for binary file reading

### LINKED ISSUE #447: UnicodeDecodeError when opening OMNIC file on Linux
I am attempting to open an OMNIC .map FTIR file, but receiving an error when I attempt to do so on my Linux machines. 

```python
from PyMca5.PyMcaIO import OmnicMap
omnic_file = OmnicMap.OmnicMap(filename)

~/.conda/envs/test_env/lib/python3.7/site-packages/PyMca5/PyMcaIO/OmnicMap.py in __init__(self, filename)
     67         else:
     68             fid = open(filename, 'r')
---> 69         data = fid.read()
     70         fid.close()
     71

~/.conda/envs/test_env/lib/python3.7/codecs.py in decode(self, input, final)
    320         # decode input (taking the buffer into account)
    321         data = self.buffer + input
--> 322         (result, consumed) = self._buffer_decode(data, self.errors, final)
    323         # keep undecoded input until the next call
    324         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xca in position 290: invalid continuation byte
```

I am not sure what is ultimately causing this error, but if I change https://github.com/vasole/pymca/blob/master/PyMca5/PyMcaIO/OmnicMap.py#L68 to open the file as a binary (like specified for when running on a Window system, the file loads with no problem).

Any ideas?

jat255:
Resolves #447 

Note: I do not have access to a Windows or Mac system to test this change, but the code logic has not changed for Windows, so that case should be fine. The only thing that should probably be tested is MacOS (`darwin`).

### DIFF
--- a/PyMca5/PyMcaIO/OmnicMap.py
+++ b/PyMca5/PyMcaIO/OmnicMap.py
@@ -62,10 +62,7 @@ class OmnicMap(DataObject.DataObject):
             It is expected to work with OMNIC versions 7.x and 8.x
         '''
         DataObject.DataObject.__init__(self)
-        if sys.platform == 'win32':
-            fid = open(filename, 'rb')
-        else:
-            fid = open(filename, 'r')
+        fid = open(filename, 'rb')
         data = fid.read()
         fid.close()
 


vasole:
Thanks!

PR #458, opened 2019-05-06
[ci] enable Qt tests

woutdenolf:
I wanted to enable GUI testing on travis (for comparing new vs. legacy PyMcaBatch code) and found that "Linux, py36, pyqt5" has leaking widgets. I haven't been able to reproduce the error locally. So use PySide2 or try to solve the issue?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,62 @@
+dist: xenial
+
 language: python
 
 matrix:
     include:
         - python: 2.7
           os: linux
-          env: BUILD_COMMAND=sdist
+          env:
+              - BUILD_COMMAND=sdist
+              - QT_BINDING=PySide2
 
         - python: 3.6
           os: linux
-          env: BUILD_COMMAND=sdist
+          env:
+              - BUILD_COMMAND=sdist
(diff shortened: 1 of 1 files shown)


vasole:
You can give a try at using PySide2 to see if there is any improvement.

I have systematic crashes when leaving  any PyQt5 application in my machine that disappear when using PySide2.

woutdenolf:
Sure, it works with PySide2 but that is probably just because widget leak testing is [disabled for pyside](https://github.com/woutdenolf/pymca/blob/3bd3e13eca5f3ac9bd187d1a93faeecf486945dc/PyMca5/PyMcaGui/misc/testutils.py#L217). Windows and Mac do not have leaking widgets with PyQt5.

As for the crashes on Windows: I don't have them on my machine (python 3.7, qt 5.12.2)
``` python
python bootstrap.py -m PyMca5.PyMcaGui.pymca.PyMcaMain
```

vasole:
What windows version? I'm on windows 7.

woutdenolf:
I also have Windows 7

Issue #460, opened 2019-05-09
[install] issue on windows


woutdenolf:
I encountered a problem with pip install from source on Windows:

```powershell
C:\projects\pymca>py -3.6 -m pip install .
C:\projects>cd ..
C:\projects>py -3.6 -m PyMca5.PyMcaGui.pymca.PyMcaMain
Traceback (most recent call last):
  File "C:\Users\denolf\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "C:\Users\denolf\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "C:\Users\denolf\AppData\Roaming\Python\Python36\site-packages\PyMca5\__init__.py", line 35, in <module>
    from PyMca5.PyMcaDataDir import PYMCA_DATA_DIR
  File "C:\Users\denolf\AppData\Roaming\Python\Python36\site-packages\PyMca5\PyMcaDataDir.py", line 35
    PYMCA_DATA_DIR = 'C:\Users\denolf\AppData\Roaming\Python\Python36\site-packages\PyMca5\PyMcaData'
                    ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

vasole:
It is really weird. I would expect to happen everywhere or nowhere. The directory is reported by Python itself. Probably one should just replace '\' by '\\' in setup.py under windows.

vasole:
Please, do nothing. I take a look tomorrow.

vasole:
Thanks a lot. To reproduce the issue I had to make a Python installation for a single user. That way the \U (from \Users instead of \P from \Program) was triggering the problem.

PR #461, opened 2019-05-09
[setup.py] DATA_DIR_FROM_SETUP and DOC_DIR_FROM_SETUP unicode error

woutdenolf:
issue #460

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -518,6 +518,24 @@ build_plotting_ctools(ext_modules)
 build_xas_xas(ext_modules)
 
 
+def stringLike(s, sexample, literal=False):
+    if literal:
+        s = s.replace('\\', r'\\')
+    if isinstance(s, type(sexample)):
+        return s
+    encodings = sys.getfilesystemencoding(), 'utf-8', 'latin-1'
+    if isinstance(s, bytes):
+        func = 'decode'
+    else:
+        func = 'encode'
+    for encoding in encodings:
+        try:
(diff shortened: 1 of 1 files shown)


vasole:
I guess a smaller change not requiring dealing with bytes or unicode would have been to just make sure an "r" was written in front of the directory string...

vasole:
Please, can you check if replacing in PyMcaDataDir.py the line ``PYMCA_DATA_DIR = 'DATA_DIR_FROM_SETUP'`` by ``PYMCA_DATA_DIR = r'DATA_DIR_FROM_SETUP'``works? I would prefer not to have to guess any encoding.

vasole:
Sorry, given the fact fisx has been using the r trick (see fisx setup.py) for quite some time, I prefer to take the conservative approach of #462

woutdenolf:
At first I thought about using raw strings as well, but doesn't this cause a problem when you actually have non-ascii characters in your path?

vasole:
I do not expect a problem since the path setting is done via the os module. If there is something to be fixed, the fisx module would also have the problem. Most of the times one passes to bytes and the transform works.

image

vasole:
The figure below illustrates the situation I'm afraid off when trying different encodings

image

woutdenolf:
I see, raw strings work just fine. As for trying different encodings: isn't that why one always tries utf8 before latin1? Anyway, problem solved. Thanks!

Issue #477, opened 2019-05-22
[CI] Batch test problem with spaces in path


vasole:
![image](https://user-images.githubusercontent.com/5636251/58180721-d917a200-7caa-11e9-9b0b-cbada8ce96b2.png)

vasole:
closed by #478

PR #478, opened 2019-05-22
[PyMcaBatch] Deal with spaces in the path of the executable

vasole:
The default installation of Python when selecting installation for All Users is "C:\Program Files"


vasole:
closes #477

PR #481, opened 2019-05-23
[GUI] Add dialog to read configuration from HDF5 or ini-like files

vasole:
I should help to read fit configurations from HDF5 files.


vasole:
@woutdenolf 

This should help you. I am hesitating to change the signature of the methods at HDF5Widget level in order to pass the parent as first argument. Previously there was no parent, but that means the dialog is not shown in the center of the calling widget.

vasole:
In addition we could also modify ConfigDict in such a way that it could handle a URI

PR #485, opened 2019-05-27
[cfg] load from hdf5 in fast, advanced and batch fit dialogs

woutdenolf:


woutdenolf:
More or less the same changes you did, with additional fastfit GUI mods.

vasole:
Thanks.

PR #486, opened 2019-05-27
[XRF Fit] Config from hdf5

vasole:
Preliminary support to test API.

This PR should work on McaAdvancedFit.py and on PyMcaBatch.py


vasole:
@woutdenolf 

Sorry, we were working on the same issue. You can take over. Please feel free to change the signature of recently introduced stuff. I guess at a certain point we'll need to complement the function PyMcaFileDialogs.getFileList with a getFileOrUriList that allows to specify input and output to HDF5 files. However, that is not a priority now.

Issue #491, opened 2019-05-28
[Batch] Logging level should be INFO for user choices


vasole:
Currently there are warnings when overwriting files or not saving. Given the fact those are user choices, the logging level should be INFO instead of WARNING.

PR #493, opened 2019-05-28
[outbuffer] overwrite warnings to info

woutdenolf:
Fix issue #491


vasole:
Missing the same correction when using nosave in the batch.

PR #499, opened 2019-05-28
Add missing test.

vasole:
One test was not run in manual mode.

### DIFF
--- a/PyMca5/tests/ConfigDictTest.py
+++ b/PyMca5/tests/ConfigDictTest.py
@@ -196,6 +196,8 @@ def getSuite(auto=True):
         # use a predefined order
         testSuite.addTest(testConfigDict("testConfigDictImport"))
         testSuite.addTest(testConfigDict("testConfigDictIO"))
+        if HAS_H5PY:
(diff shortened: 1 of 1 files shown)


PyMca5/tests/ConfigDictTest.py
```diff
@@ -196,6 +196,8 @@ def getSuite(auto=True):
         # use a predefined order
         testSuite.addTest(testConfigDict("testConfigDictImport"))
         testSuite.addTest(testConfigDict("testConfigDictIO"))
+        if HAS_H5PY:
```
woutdenolf: Perhaps use `unitests.skipIf` for this kind of thing?

vasole: That is the approach used for the automatic testing (it is there in line 125).

It was just missing when running the test standalone.

woutdenolf: I see, but then why not rely on that and not exclude it explicitly here?

vasole: Thanks. I misunderstood how it was working.

Issue #489, opened 2019-05-28
[Batch Fit] Chisquare image not generated


vasole:
The chisq image is missing among the generated EDF files not inside the .dat file

woutdenolf:
This is only saved when `diagnostics` is enabled. Should I always save it? The `OutputBuffer` has a `diagnostics` property which enables/disables `saveFOM`, `saveResiduals`, `saveData`, `saveFit`. The first stands for figures-of-merit (currently contains chisq, nFreeParameters, nObservations). What I could do is make sure `saveFOM` is always set, no matter what the user selects in the GUI. But then you get all FOMs. Should I consider chisq separately?

woutdenolf:
No sorry, that's for HDF5. For EDF and DAT, only the chisq is saved from the FOMs. But still only when `diagnostics` is enabled by the user.

vasole:
In the past it was systematically saved.

vasole:
closed by #500

PR #500, opened 2019-05-28
[batchfit] force save FOMs (chisq only for non-hdf5 formats)

woutdenolf:
Solve issue #489. The FOMs will always be saved. HDF5 will contain all of them, the other formats only the chisq. Note that this only goes for slow fitting, fast fitting does not produce the chisq.


vasole:
Wow! I had not realized that the fast fitting was not producing any figure of merit.

Issue #422, opened 2019-02-08
[GUI] Read images from HDF5 files


vasole:
PyMca relies on EDF to read image data at several places. It is critical to be able to do the same from HDF5 files at least in the External Images plugins and the RGB correlator.

woutdenolf:
As for the external images plugin, should I do something like this? @vasole 
```
class ExternalImagesStackPlugin(StackPluginBase)

    def _loadImageFiles(self):
        # prompt for files
        for filename in filelist:
            edf -> EdfFileDataSource -> StackPluginResultsWindow
            dat, csv -> specfilewrapper -> StackPluginResultsWindow
            h5 -> HDF5Stack1D -> StackPluginResultsWindow
            else -> QImage -> ExternalImagesWindow
```

vasole:
- I fail to catch the need for the prompt keyword. I guess you can use HDF5Widget.getURI and accept URIs as filenames for single datasets and groups. Given the fact the RGB Correlator is a graphical application, it can always prompt for an URI in case of receiving just an HDF5File
- I would avoid the use of HDF5Stack1D for the external images. I find that class somewhat fragile. Is it really adapted to StackPluginResultsWindow? The later expects lists of images and not a stack of images. For browsing a stack, the StackBrowserPlugin uses StackBrowser that is more appropriate for browsing stacks.

vasole:
Concerning the refactoring, I would avoid it and work just adding the HDF5 support with the minimal changes. Hopefully it will be the last added format.

vasole:
Basically in addFileList one needs to check 

- if the file exists and if it is an HDF5 file to open and additional dialog to fetch the URI
- if the file does not exist check if we have got an URI

woutdenolf:
I added HDF5 support to the RGB correlator and External Images Plugin. Any others?

PR #502, opened 2019-05-29
[rgb] add hdf5 support

woutdenolf:
This PR addresses issue #422 (only the RGB correlator).

I minimized refactoring:
 - `RGBCorrelatorWidget.addFileList` loops and calls `_addSingleFile` which checks file formats and calls `_addEdfFile`, `_addHdf5File`, etc.
 - remove file format checking from `PyMcaPostBatch` (redundant, didn't want to do it for HDF5 as well)
 - deprecation warning for the public `RGBCorrelatorWidget.addBatchDatFile` method (calls `addFileList`)

HDF5 support is added to `RGBCorrelatorWidget`:
 - If the file exists but the uri does not, it will prompt for a uri.
 - I used `HDF5Widget.getUri` and `NexusUtils.getNdimDatasets`.

### DIFF
--- a/PyMca5/PyMcaGui/pymca/PyMcaMain.py
+++ b/PyMca5/PyMcaGui/pymca/PyMcaMain.py
@@ -1224,11 +1224,7 @@ class PyMcaMain(PyMcaMdi.PyMcaMdi):
             correlator.raise_()
         self.__correlator[-1].sigRGBCorrelatorSignal.connect( \
                 self._deleteCorrelator)
-
-        if len(filelist) == 1:
-            correlator.addBatchDatFile(filelist[0])
-        else:
-            correlator.addFileList(filelist)
+        correlator.addFileList(filelist)
 
     def __sumRules(self):
         if self.__correlator is None:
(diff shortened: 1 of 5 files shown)


vasole:
Usually we are not forcing the number of dimensions but the size. If the size is correct, the RGBCorrelator can resize the dataset. The use case is to read the I0 (or some other monitor) in order to divide/correct the output of a fit.

vasole:
For instance, in some widgets to load motor positions, we accept size equal 1 or size equal to image size despite the shape not being the same.

vasole:
I forgot to add that if the user selects a Group, then the compatible (in size!) datasets present in the group are selected.

woutdenolf:
What shape do we take as a reference when loading the first HDF5 group? If it's an NXdata, we can take the signal shape. But if not? The one with the largest dimensionality? The one with the most occurrences?

vasole:
Unless all the datasets in a group have the same shape, I would force the user to select a dataset and not a group.

woutdenolf:
Changes concerning data size.

When loading HDF5 datasets, the following restrictions are applied:
 - when NXdata instance: datasets must be marked as signals (including aux)
 - when RGBcorrelator has data: datasets must have the same number of elements (reshaped in `addImage` as usual)
 - when RGBcorrelator has no data: datasets must be 1D or 2D (causes problems down the line otherwise)

Datasets that do not match these restrictions are silently ignored. The user is notified only when no dataset matches or when matching datasets have different sizes (popup warning + relaunch the uri selection gui).

PR #506, opened 2019-05-30
[extimg] add hdf5 support

woutdenolf:
This PR addresses issue #422 (only the External Images Plugin). Same dataset selection logic as RGB Correlator (PR #502). Except in this case, the data shape seems to be restricted instead of the number of elements (same shape, but allowing for transposition). At least this is the behavior for EDF files so I assume it is also expected for HDF5 datasets.


vasole:
Silx External Images Tool also needs to access the HDF5 results.

vasole:
It is another Stack plugin.

woutdenolf:
Done. IO routines are shared between the 2 extimg plugins.

vasole:
Currently re-running the tests under MacOS to see if the failure is reproducible.

vasole:
I dislike the use of a metaclass but well...

I think ExternalImagesStackPluginBase.py should be moved to PyMca5.PyMcaGui.pymca

PR #526, opened 2019-07-05
Branch for 5.5.0

vasole:


woutdenolf:
All tests are now passing on my win7/py27.

vasole:
On my machine the testSlowRoiFitEdfMap and the testSlowRoiFitSpecMap do not pass.

Perhaps we should consider a separate issue for them (PyQt4 should be dropped in any case).

Issue #510, opened 2019-06-11
[ROI Batch] Support command line and HDF5


vasole:
The ROI batch should evolve in an analogue way to the FastXRFLinearFit in the sense of accepting HDF5 files from the command line and generating HDF5 output.

woutdenolf:
So basically this means using the `OutputBuffer` in `StackROIBatch`? Does it need to be done before the next release? And when is that release?

vasole:
I do not think it is necessary for 5.5.0 (but certainly for 5.5.1).

I have to solve #479 and deal with #466 in order to make the 5.5.0

The problem is that silx 0.10.1 is already in debian experimental and that breaks PyMca.

vasole:
Closed by #528

PR #528, opened 2019-07-05
[roibatch] add hdf5 support

woutdenolf:
Implement feature request #510.

GUI and CLI options are improved to look like the fast fitting UI:

image

Implementation details: modified StackROIBatch to re-use two XRF fitting utils
 1. OutputBuffer (mem alloc and saving): extracted the generic part (non-xrf fit specific) to PyMca5.PyMcaIO
 2. McaStackView (iterate over MCA stack in chunks): moved to PyMca5.PyMcaCore


woutdenolf:
@vasole This was done for a while but I wanted to do some manual testing on top of the unittests. That's done now and everything seems fine.

vasole:
I'm just wondering about the move of McaStackView to PyMcaCore. Somehow that later container is for things fairly specific to PyMca and/or where I consider I have more "freedom"

Perhaps the one that was misplaced was the StackROIBatch that is just a math operation not really specific to PyMca?

BTW, with McaStackView you introduce a dependency on six. Unlike the colleagues developing the silx library, in PyMca I preferred to be explicit about what was Python2 and what Python3. 

Pure number crunching (=no GUI) on PyMca only requires numpy and eventually h5py for I/O. So, probably is better to put McaStackView in PyMcaCore after all because I want to to get rid of that dependency.

I have seen that you also used six in XRFBatchFitOutput

vasole:
I will open a new issue to get rid of six.

PR #529, opened 2019-07-09
Remove unused variables

payno:
This variable allocation is raising a KeyError in my case when try to access `k`. Without processing seems find.
Anyway those variables are never used later (or they are overwrite)


vasole:
What is k in your case?

payno:
It appends when I call XASClass.processSpectrum with a specific spectrum from the 'S5_nuit1_merge_new_crop.h5' dataset.
So those are supposed to be 'EXAFSNormalized' / 'normalized mu'.
It does not appends for each spectrum.

PR #550, opened 2019-08-01
[CI] Test python 3.7 instead of 3.6

vasole:


vasole:
It is better to keep testing 2.7, 3.6 (linux) because 3.7 is already tested on MacOS.

PR #556, opened 2019-08-06
Do not lock HDF5 files if used as main application.

vasole:


vasole:
Should the batch fitting avoid locking the files? For reading it should not lock them but the doubt is for writing them.

PR #560, opened 2019-08-08
[RegularMeshPlugin] Deal with bliss way of calculating the coordinates.

vasole:
Difference of one point when comparing to SPEC


vasole:
It seems bliss is going to follow the SPEC convention. Then this modification would fail again.

PR #567, opened 2019-08-12
Remove explicit six dependency

### LINKED ISSUE #566: Get rid of six explicit dependency
Used in ExternalImagesStackPluginBase, McaStackView, XRFBatchFitOutput, OutputBuffer

vasole:
closes #566

### DIFF
--- a/PyMca5/PyMcaCore/McaStackView.py
+++ b/PyMca5/PyMcaCore/McaStackView.py
@@ -33,8 +33,6 @@ __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
 
 import numpy
 import logging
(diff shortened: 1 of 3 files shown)


vasole:
@woutdenolf 

I did not find a clean way to match the syntax of Python 2 and Python 3 with metaclasses and the six with_metaclass is far from providing easy to understand code.

Python 2

```
class MyClass(object):
    __metaclass__ = Meta
```
Python 3

```
class MyClass(metaclass=Meta):
    pass
```

ChunkedView is not inherited outside the module containing it, so, there is no big deal with the change.

Only concern could be ExternalImagesStackPluginBase that is inherited at one place.

Issue #603, opened 2019-10-13
[HDF5] Handle broken NXentry links


woutdenolf:
`HDF5Stack1D` dataset selection gives following errors when there are broken NXentry links:

```bash
Traceback (most recent call last):
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaGui/pymca/PyMcaMain.py", line 1444, in __roiImaging
    widget.loadStack()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaGui/pymca/QStackWidget.py", line 581, in loadStack
    stack = self.stackSelector.getStack()
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaGui/pymca/StackSelector.py", line 106, in getStack
    stack = QHDF5Stack1D.QHDF5Stack1D(filelist)
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaGui/pymca/QHDF5Stack1D.py", line 51, in __init__
    dtype=dtype)
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaIO/HDF5Stack1D.py", line 67, in __init__
    self.loadFileList(filelist, selection, scanlist)
  File "/mntdirect/_data_id21_inhouse/wout/dev/pymca/build/lib.linux-x86_64-3.6/PyMca5/PyMcaIO/HDF5Stack1D.py", line 97, in loadFileList
    if isinstance(tmpHdf["/"+key], h5py.Group):
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/mntdirect/_data_id21_inhouse/wout/dev/virtualenvs/rnice8/pymca/py36/lib/python3.6/site-packages/h5py/_hl/group.py", line 262, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
[...]

PR #604, opened 2019-10-13
Handle broken NXentry links

woutdenolf:
Solved #603


vasole:
I have merged it into master. We should provide some logging when that happens (INFO level at the very least).

PR #611, opened 2019-11-09
XRFMCHelper: use GTK instead of GTK2

tschoonj:
This has been GTK for +3 years now...


vasole:
OK. I am going to be a bit conservative. I will check if the GTK2 directory exists and if not I will default to the new version. Some people are still using the XMI-MSIM 3.1 version it was shipped with the training.

tschoonj:
Thanks Armando!

Issue #641, opened 2020-01-17
[PCA] Give the user the choice to standardize the data


vasole:
Offer the choice between performing PCA on the covariance matrix or the correlation matrix. The later should work better when combining different techniques (option PCA on multiple arrays).

PR #645, opened 2020-02-03
[PCA] Allow to use covariance or correlation

vasole:
Give the user the choice to perform the PCA on the covariance matrix or the correlation matrix, the later being the covariance matrix scaled to the diagonal elements in order to have unit standard deviation.


vasole:
closes #641

Issue #89, opened 2017-09-11
[ROI Imaging] Read and write scales and calibrations.


vasole:
Information  read from some formats like JCAMP-DX is lost when exporting the stack.

Clearly the internal model used by PyMca should be exported as an HDF5 file.

vasole:
#692 saves actual motor positions, but not scales.

vasole:
closed by #692 and #699

PR #699, opened 2020-05-17
[ROI Imaging] Save and load scales if present

vasole:
Allow to save and to retrieve stack, calibration, time, positiones and scales (if any) in such a way that information can be retrieved selecting a single dataset.


vasole:
closes #89

PR #705, opened 2020-07-21
Add note about OpenGL

mpmdean:
@vasole I would suggest including this note.


vasole:
Mesa is certainly not a dependency of PyMca. On Windows or Mac, the compiling environment provides the OpenGL headers by default and Mesa is not needed. Debian/Ubuntu do not provide those headers by default. It is more an issue of completeness of the compiling enviroment. I will try to find other solution to test if OpenGL programs can be compiled and, if it is not the case, skip the compilation of the OpenGL related modules.

mpmdean:
Sounds good

PR #710, opened 2020-08-11
[fastfit] add peak profiles to diagnostics

woutdenolf:
Useful for several things, but mostly to debug bad fits.


vasole:
There is a risk with saving the channels as int32. Sometimes people provide the actual energy instead of the channels with a (0, 1, 0) calibration.

PR #711, opened 2020-08-11
[WIP] multidetector xrf fits

woutdenolf:


vasole:
I think it would be much less intrusive to do the following:

- ClassMcaTheory.py becomes LegacyClassMcaTheory
- The new ClassMcaTheory imports McaTheory from LegacyClassMcaTheory
- The new Class (if any) in ClassMcaTheory is named ClassMcaTheory as it is the style followed everywhere else in the project.

woutdenolf:
> ClassMcaTheory.py becomes LegacyClassMcaTheory.py

That's already the case.

> The new Class (if any) in ClassMcaTheory is named ClassMcaTheory

Yes that will be the case.

> The new ClassMcaTheory imports McaTheory from LegacyClassMcaTheory

That's not going to be possible imo.

vasole:
You are correct. At a certain point I put a dummy ClassMcaTheory to follow the convention available everywhere else :(

vasole:
We'll have to take care.

There are recipes around showing how to directly use the current McaTheory "public" API (configure, setData, estimate, startfit)

PR #718, opened 2020-09-15
fix fwhm calculation

dschick:
when doing the simple spec arithmetic I encountered a bug in the math which is only obvious for data with a few (1-2) points at the slope (plateau-like instead of gauss-like functions).

for finding the x-coordinates of the FWHM a `while` loop is run for the lower and upper side of the peak maximum to find the index of the first data point that is smaller than the half-max value.

in the subsequent linear interpolation one should then use one data point below and above the half-max value.
This PR fixes this issue for the upper side of the peak.

### DIFF
--- a/PyMca5/PyMcaMath/SpecArithmetic.py
+++ b/PyMca5/PyMcaMath/SpecArithmetic.py
@@ -102,10 +102,10 @@ def search_fwhm(xdata,ydata,peak=None,index=None):
         while ydata[idx] >= hm:
             idx = idx+1
 
-        x0 = xdata[idx]
-        x1 = xdata[idx+1]
-        y0 = ydata[idx]
-        y1 = ydata[idx+1]
+        x0 = xdata[idx-1]
+        x1 = xdata[idx]
+        y0 = ydata[idx-1]
+        y1 = ydata[idx]
 
         uhmx = (hm*(x1-x0) - (y0*x1)+(y1*x0)) / (y1-y0)
     except ZeroDivisionError:


dschick:
just another side note:
SPEC automatically notices if the data is peak- or erf-like and calculates in both cases a meaningful FWHM and CFWHM.

We recently added this functionality to [Sardana](https://github.com/dschick/sardana/tree/calcScanStats) by simply calculating the derivative of erf-like data before calcuating FWHM and CFWHM. 

I was wondering, if this might be alo interesting for PyMCA?

vasole:
Thanks.

Concerning your question, the way to calculate the derivative influences the resulting FWHM: smoothing, not smoothing, number of points and noise lead to very different results in case of very sharp steps (or poorly sampled ones).

The approach to follow is to directly fit the data with the correct modelling function. Those "SPEC Arithmetic" functions are there to mimic what SPEC does and not to provide actual reliable results. If you want correct FWHM values, do a fit to the raw data (no smoothing, no derivative) and forget about those "SPEC Arithmetic" calculations.

dschick:
thanks for your quick reply.
Sure one sould use a fit function for best/reliable results.

Here, we are talking more about practical helpers when aligning an experiment and doing knife-edge or pinhole scans.

I was just asking about the error-funtion-like or step-like situation, because PyMCA does not mimic SPEC here correctly.
SPEC does recognize if the data is step- or peak-like and calculates the CFWHM "correctly" for both cases.

vasole:
Then feel free to make a PR. Please keep in mind it the calculation should not be "hungry" on resources.

PR #719, opened 2020-09-15
automatic step-like data detection

dschick:
check if the data is peak- or step-like.

Calculate the gradient of step-like data and recalculate the half-max value


vasole:
I have merged it, but there are weird things like negative FHWM happening.

We'll see if there are complaints. I never took those numbers seriously.

PR #734, opened 2020-09-21
Rgb correlator apriori shape

### LINKED ISSUE #729: [RGB] scatter plot
When doing a 2D XRF map with ROI counters (saved as flat datasets) it is currently not possible to open them in RGB correlator.

woutdenolf:
Closes #729 

Add image shape to `pymcapostbatch` CLI:

```python
pymcapostbatch --shape=68,101 filename.h5::1.1/measurement
```

In addition the RGB correlator window is opened when no images are specified. Useful in case you want to specify the shape (with the "reshape" button) before loading images when launching the RGB correlator from the main GUI.

### DIFF
--- a/PyMca5/PyMcaGui/pymca/PyMcaPostBatch.py
+++ b/PyMca5/PyMcaGui/pymca/PyMcaPostBatch.py
@@ -131,11 +131,11 @@ def main():
     w.layout().setContentsMargins(11, 11, 11, 11)
     if not filelist:
         filelist = w._getStackOfFiles()
-    if not filelist:
+    if filelist:
+        w.addFileList(filelist)
+    else:
         print("Usage:")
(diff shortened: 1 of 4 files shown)


PyMca5/PyMcaGui/pymca/PyMcaPostBatch.py
```diff
@@ -119,19 +120,22 @@ def main():
         elif opt in '--fileindex':
             if int(arg):
                 transpose = True
+        elif opt in '--shape':
+            image_shape = tuple(int(n) for n in arg.split(','))
```
vasole, line 124: It seems more logical to pass the shape as n_rowsxn_columns than as nrows,ncolumns. I think ImageMagick uses that approach.

PR #803, opened 2021-04-08
Verify risk of overwriting input file when choosing output directory

### LINKED ISSUE #800: Crash on batch fit using input directory as output directory
image

vasole:
The risk appears when there is only one HDF5 file to fit and the output directory is the directory containing that file.

closes #800 
completes PR #802


vasole:
image

PR #821, opened 2021-12-09
fix: protect `HDF5Widget.emitSignal()` when mouse was not pressed yet

reszelaz:
`_lastMouse` attribute does not exist if the mouse was never pressed in the widget lifetime. This causes troubles e.g. when one sets current index (`setCurrentIndex()`) which leads to `emitSignal()`.
Protect `emitSignal()` against non existing `_lastMouse` attribute.

### DIFF
--- a/PyMca5/PyMcaGui/io/hdf5/HDF5Widget.py
+++ b/PyMca5/PyMcaGui/io/hdf5/HDF5Widget.py
@@ -827,10 +827,7 @@ class HDF5Widget(FileView):
         ddict['dtype'] = item.dtype
         ddict['shape'] = item.shape
         ddict['color'] = item.color
-        mouse = None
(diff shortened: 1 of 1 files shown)


vasole:
Thanks.

I wonder if one should set the mouse key to "left" to be consistent with line 786 instead of setting it to None. Have you tried that possibility?

reszelaz:
Yes, it also works with `"left"`. 

Actually, this way of solving it is shorter:

```python
ddict['mouse'] = getattr(self, '_lastMouse', 'left') * 1
```
If you prefer this one, I will apply this change.

vasole:
Thanks. Please do so. I do not know if setting it to None can give troubles somewhere else.

reszelaz:
Ok. Done.

PR #844, opened 2022-03-30
[GUI] Get rid of .exec_ calls

### LINKED ISSUE #843: [PyMcaPostBatch] HDF5 dialog does not show anything
```
pymcapostbatch /path/to/file.h5
```
image

Some quick debugging shows that `Hdf5SelectionDialog.exec_` is not called, neither by PyQt5 5.9.2 nore 5.15.6. The difference was that the 5.15.6 installation was done from source which has  commit https://github.com/vasole/pymca/commit/6da6ab52efe44e3eb3caa3efd4d42ff9b1a344e8

https://github.com/vasole/pymca/blob/871ff3cef08cf5d5fd0ba589be289c56eb3e0de3/PyMca5/PyMcaGui/io/hdf5/HDF5Widget.py#L956-L963

So I guess it's fixed but we need to do the same for all our widget classes where we overwrite `exec_`.

vasole:
Should close #843 alike issues.

### DIFF
--- a/PyMca5/PyMcaGui/PluginsToolButton.py
+++ b/PyMca5/PyMcaGui/PluginsToolButton.py
@@ -1,8 +1,8 @@
 #/*##########################################################################
-# Copyright (C) 2004-2019 V.A. Sole, European Synchrotron Radiation Facility
+# Copyright (C) 2004-2022 European Synchrotron Radiation Facility
 #
 # This file is part of the PyMca X-ray Fluorescence Toolkit developed at
-# the ESRF by the Software group.
+# the ESRF.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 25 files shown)


woutdenolf:
Could you explain the logic of this change?

1. We replace all `someqtobject.exec_()` with `someqtobject.exec()` which means we drop support of older pyqt/pyside versions. Do you know from which versions? Shouldn't this be reflected in `setup.py` as well?
2. We have this pattern
    ```
    def exec_(self):
        return self.exec()

    def exec(self):
        ...
    ```
    which I guess is meant to support both old and new qt versions. This seems to contradict the first change.

vasole:
It is already mentioned. Now PyMca supports PyQt5, PySide2 and PySide6.

To keep the exec_ method is for people using PyMca as a library.

woutdenolf (approved):
Ok thanks. LGTM.

PR #846, opened 2022-03-30
[NexusTools] Get scanned positioners account for special scans (fscan for the time being)

vasole:
Some scans have the "scanned" positionner configured as a counter.

This modification will show "scanned" counters as first in the list for particular scans.


vasole:
@woutdenolf An alternative to this approach would be that the scan inform the writer about the counter to be considered as positioner so that it can be added to the list of positioners.

For the time being I would leave it as it is.

PR #847, opened 2022-03-31
[HDF5] optionally exclude the last scan from an HDF5 model tree

woutdenolf:
imageimage


vasole:
Unless I have misunderstood it, it seems the PR tries to get the items always using a subprocess. If that is the case, I would prefer to first check if the file is older than one minute to use the old way in that case. It seems overkill to always open a subprocess.

woutdenolf:
Yes it's only for the root.

vasole:
PR https://github.com/vasole/pymca/pull/848 made this not necessary

PR #848, opened 2022-03-31
[HDF5] Hdf5 segfault handling

woutdenolf:
Iterating the HDF5 root (for example `h5file["/"].keys()`) can cause segfaults. Execute these calls in a subprocess.


vasole:
I have added a condition for the file to be older than 1 hour.

PR #869, opened 2022-06-09
Clean up locale_management in specfile

### LINKED ISSUE #859: ‘LOCALE_NAME_MAX_LENGTH’ undeclared
I tried to `python3 -m pip install .` the latest version of PyMca on Centos 7 but building failed with

```
 PyMca5/PyMcaIO/specfile/src/locale_management.c: In function ‘PyMcaAtof’:
    PyMca5/PyMcaIO/specfile/src/locale_management.c:44:20: error: ‘LOCALE_NAME_MAX_LENGTH’ undeclared (first use in this function)
      char localeBuffer[LOCALE_NAME_MAX_LENGTH + 1] = {'\0'};
                        ^~~~~~~~~~~~~~~~~~~~~~
    PyMca5/PyMcaIO/specfile/src/locale_management.c:44:20: note: each undeclared identifier is reported only once for each function it appears in
    PyMca5/PyMcaIO/specfile/src/locale_management.c:45:9: error: redeclaration of ‘result’ with no linkage
      double result;
             ^~~~~~
    PyMca5/PyMcaIO/specfile/src/locale_management.c:30:9: note: previous declaration of ‘result’ was here
      double result;
             ^~~~~~
    PyMca5/PyMcaIO/specfile/src/locale_management.c:44:7: warning: unused variable ‘localeBuffer’ [-Wunused-variable]
      char localeBuffer[LOCALE_NAME_MAX_LENGTH + 1] = {'\0'};
           ^~~~~~~~~~~~
    PyMca5/PyMcaIO/specfile/src/locale_management.c:43:8: warning: variable ‘restoredLocaleBuffer’ set but not used [-Wunused-but-set-variable]
      char *restoredLocaleBuffer;
            ^~~~~~~~~~~~~~~~~~~~
    error: command 'gcc' failed with exit status 1
```

The same happens when running python3 -m pip install PyMca5.

llohse:
Harmonize use of strtod_l. It should be available on all supported OSes.
Makes SPECFILE_USE_GNU_SOURCE obsolete.

fixes #859

### DIFF
--- a/PyMca5/PyMcaIO/specfile/include/locale_management.h
+++ b/PyMca5/PyMcaIO/specfile/include/locale_management.h
@@ -24,23 +24,6 @@
 #define PyMca_LOCALE_MANAGEMENT_H
 
 #include <locale.h>
-#ifdef _GNU_SOURCE
-#  ifdef __GLIBC__
-#    include <features.h>
-#    if !((__GLIBC__ > 2) || ((__GLIBC__ == 2) && (__GLIBC_MINOR__ > 25)))
-#      /* strtod_l has been moved to stdlib.h since glibc 2.26 */
-#      include <xlocale.h>
-#    endif
-#  else
-#    include <xlocale.h>
-#  endif
-#else
-#  ifdef SPECFILE_POSIX
-#    ifndef LOCALE_NAME_MAX_LENGTH
-#           define LOCALE_NAME_MAX_LENGTH 85
-#    endif
-#  endif
-#endif
 
 double PyMcaAtof(const char*);
 
--- a/PyMca5/PyMcaIO/specfile/src/locale_management.c
+++ b/PyMca5/PyMcaIO/specfile/src/locale_management.c
@@ -29,7 +29,7 @@ double PyMcaAtof(const char * inputString)
     double result;
 	_locale_t newLocale = _create_locale(LC_NUMERIC, "C");
 	result = _atof_l(inputString, newLocale);
-	_freelocale(newLocale);
+	_free_locale(newLocale);
 	return result;
 #else
 	double result;
--- a/PyMca5/PyMcaIO/specfile/src/locale_management.c
+++ b/PyMca5/PyMcaIO/specfile/src/locale_management.c
@@ -20,6 +20,9 @@
 # THE SOFTWARE.
 #
 # ############################################################################*/
+
+#define _GNU_SOURCE
+
 #include <stdlib.h>
 #include <locale_management.h>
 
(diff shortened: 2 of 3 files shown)


vasole:
As it is now, when __USE_GNU is not defined, the dependency of atof on the locale will make things not work. Is __USE_GNU defined in *all* POSIX systems?

I was considering to implement a default testing the returned value of `atof("2.5")` and if not being 2.5, then implement a new atof where the dot, if present is replaced by a `,` and atof called.

llohse:
> Also, it is needed to include xlocale.h for GLIBC versions prior to 2.26:
> 
> ```
> #  if !((__GLIBC__ > 2) || ((__GLIBC__ == 2) && (__GLIBC_MINOR__ > 25)))
> #      /* strtod_l has been moved to stdlib.h since glibc 2.26 */
> #      include <xlocale.h>
> #  endif
> ```

According to this site [1], you don't. xlocale.h was never intended to be used anywhere outside of glibc itself. Supposedly, the definitions have always been available thorugh stdlib.h.
I checked on a Centos 6.10 with glibc 2.12, which had the proper definitions in stdlib.h

[1] https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27

llohse:
> As it is now, when __USE_GNU is not defined, the dependency of atof on the locale will make things not work. Is __USE_GNU defined in _all_ POSIX systems?

Good question. But I guess it is enough to support the platforms supported by Python itself? 

Since currently the compiler is called by setuptools, the question is whether setuptools disables the GNU extensions on any supported platform.

How do you feel about hardcoding `#define _GNU_SOURCE 1` in locale_management.c to be on the safe side? In that case we could even exclude the test for __USE_GNU and just assume that it is either Windows or provides the necessary GNU extensions. Also, in that case compilation would fail if the functions are not available - which would immediately show if there is a problem.

In my opinion the cleanest and best way would be to use a proper build system like CMake or Meson to test if the functions are available and take care of all the necessary compiler flags. There one could easily set the C version to gnu99 and be done with it. But that is a different story.

> I was considering to implement a default testing the returned value of `atof("2.5")` and if not being 2.5, then implement a new atof where the dot, if present is replaced by a `,` and atof called.

Sounds good to me. But it would not be necessary if one just had to distinguish between either Windows or a system with GNU extensions.

vasole:
The function strtod_l does not seem to be present with the Intel C compiler.
I think it is best to check for the presence of the GNU compiler in the same way I do for the presence of the Microsoft compiler. That way the code can be working as before while not needing any input flag.

llohse:
> The function strtod_l does not seem to be present with the Intel C compiler. I think it is best to check for the presence of the GNU compiler in the same way I do for the presence of the Microsoft compiler. That way the code can be working as before while not needing any input flag.

That is weird. It *should* not be a compiler issue since the function is provided by libc and not gcc...

llohse:
Yes. It is an extension from GNU libc and not provided by POSIX libc. But you should be able to link against GNU libc also from icc without any trouble. 

In any case. Your check for __GLIBC__ looks reasonable to me.

PR #906, opened 2022-09-22
Fix rounding within OmnicMap's getPositionFromIndexAndInfo

markotoplak:
This PR aims to fix an intermittent issue in reading positions from maps.

@clsandt sent me a map where:
- deltaX = 49.97058868408203
- x0 = -815.0
- x1 = 884.0

The number of jumps (intervals) in X is ((x1 - x0) / deltaX) which is 33.9999996946446, which means that the number of measurements nX is 33.9999996946446 + 1.

The old code wrongly rounded this down. This is what the consequences look like:
before

The new code rounds before converting to int so it is more robust to numeric inaccuracies. I think the change is self-explanatory. After the change (also as Omnic's software shows it):

after

### DIFF
--- a/PyMca5/PyMcaIO/OmnicMap.py
+++ b/PyMca5/PyMcaIO/OmnicMap.py
@@ -309,7 +309,7 @@ class OmnicMap(DataObject.DataObject):
         x1, y1 = ddict['Last map location']
         deltaX = ddict['Mapping stage X step size']
         deltaY = ddict['Mapping stage Y step size']
-        nX = int(1 + ((x1 - x0) / deltaX))
+        nX = int(round(1 + ((x1 - x0) / deltaX)))
         x = x0 + (index % nX) * deltaX
         y = y0 + int(index / nX) * deltaY
         return x, y


vasole:
Something is wrong then. I use int(1 + x) to make sure to use mathematical rounding. In python 3.x round returns banker's rounding that is not desirable. Mathematical rounding `round(x) = int(1 + x)`

Your PR would imply that the correct way to implement things would be 1 + int(1 + (x1-x0) /deltaX)) Do you have other example datasets?

markotoplak:
No, mathematical rounding is not `round(x) = int(1 + x)`. Counterexamples: `x = 3.4`, `x = 3`.

Are you sure you are not using `1 + x` to get the correct number of points from the number of intervals, which is computed by division? The old code worked correctly when the division returned exact integers of some slightly larger. I can not see how the old code could be correct if division was even slightly inaccurate into the other direction.

markotoplak:
And I doubt we'd need to take care of the banker's rounding here: the number of intervals should get very close to the correct value, like here, where the difference is less than `1e-6`.

markotoplak:
Now I remember, you probably meant `int(x+0.5)` for rounding, but that is different that was being used here and I'd still rather use python's `round()` here.

If you are more comfortable with `int(x+0.5)`, I could change `round` to that, because I think there should be no difference.

vasole:
Sorry, my mistake. I indeed meant int(x + 0.5). Please replace the use of round and it is fine with me.

markotoplak:
Thanks for merging this!

vasole:
@clsandt 

What version of Omnic was used to generate those files? I cannot extract the shape information from them and one has to reshape the map. The first one is clearly an 18x18 map, but the second is less clear.

Can you post one of the files correctly read?

clsandt:
That was version 8.1.210 of Omnic and Atlus. 
One file was indeed 18x18, the other was 14x15 (220701-10).
For both files, the images should look like a ring with material mostly on the side and little material inside.

PR #914, opened 2022-10-04
[McaAdvancedFitBatch] Account for the case the initial X values are energy values

vasole:
Sometimes the user provides energy values and a calibration (0, 1, 0) or (0, 0.001, 0).

One cannot assume the x values are integers.

### DIFF
--- a/PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
+++ b/PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
@@ -956,8 +956,8 @@ class McaAdvancedFitBatch(object):
             xdata = self.mcafit.xdata.flatten().astype(numpy.float32)   # x (after fit limits)
             stackShape = self.__nrows, self.__ncols, len(xdata0)
             mcaIndex = 2
-            iXMin, iXMax = int(max(numpy.where(xdata0 <= xdata[0]))),
-                           int(min(numpy.where(xdata0 >= xdata[-1]))) + 1
+            iXMin, iXMax = int(max(numpy.where(xdata0 <= xdata[0])[0])),
+                           int(min(numpy.where(xdata0 >= xdata[-1])[0])) + 1
             self._mcaIdx = slice(iXMin, iXMax)
             nObs = iXMax-iXMin
             if outbuffer.saveFOM:
--- a/PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
+++ b/PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
@@ -2,10 +2,10 @@
 #
 # The PyMca X-Ray Fluorescence Toolkit
 #
-# Copyright (c) 2004-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2022 European Synchrotron Radiation Facility
 #
 # This file is part of the PyMca X-ray Fluorescence Toolkit developed at
-# the ESRF by the Software group.
+# the ESRF.
 #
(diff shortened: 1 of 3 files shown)


vasole:
@woutdenolf 

Ready to merge from my side.

PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
```diff
@@ -1002,10 +1003,10 @@ def _allocateMemoryFit(self, result, concentrations):
                     #gain = result['fittedpar'][result['parameters'].index('Gain')]
                     zero = mcacfg['zero']
                     gain = mcacfg['gain']
-                    xenergy = zero + gain*xdata0
+                    xenergy = zero + gain * xdata0
```
woutdenolf, line 1006: This doesn't make sense if `xdata` can be either channels or energy.

PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
```diff
@@ -1002,10 +1003,10 @@ def _allocateMemoryFit(self, result, concentrations):
...
                     dataAxesNames[mcaIndex] = 'energy'
                     dataAxes[mcaIndex] = 'energy', xenergy.astype(dtypeResult), {'units': 'keV'}
-                    dataAxes.append(('channels', xdata0.astype(numpy.int32), {}))
+                    dataAxes.append(('channels', xdata0.astype(numpy.float32), {}))
```
woutdenolf, line 1009: This doesn't make sense if xdata can be either channels or energy.

woutdenolf (commented):
If mcafit.xdata can be either energy or channels, there will be many issues (I pointed out 2 of them but I suspect there are many more). It is probably better do deal with energy/channels in `McaTheory` (convert enery+calibration to channels)

PyMca5/PyMcaPhysics/xrf/McaAdvancedFitBatch.py
vasole: That depends on the user.

It can pass energy and then the calibration be (0, 1, 0).

woutdenolf: I suggest to deal with that in `McaTheory`. Keeping the energy/channel duality make all the rest more complicated.

vasole: Why? Dealing with it in McaTheory would imply having to change things also at the user side.

That duality has been there since the beginning. I put you as reviewer to make sure you were aware of it.

Nothing ever prevented the user from selecting an energy axis and a counts signal and request a fit. Therefore that has to be supported with no further modification.

woutdenolf: > Nothing ever prevented the user from selecting an energy axis and a counts signal and request a fit. Therefore that has to be supported with no further modification.

Indeed and I suggest supporting that by dealing with it in the `McaTheory` class.

I understand your pov and I gave you mine. Up to you.

woutdenolf: If you merge, make sure that there aren't other places like this (e.g. in OutputBuffer and fast fitting)

vasole: It seems on fast fitting you properly dealt with it in _fitMcaTrimInfo

PR #965, opened 2023-03-01
Swap arg places in nxClassNeedsInit error message

loichuder:
When trying to save fit results in `/`, I got the following error:

```
RuntimeError: / is an instance of NXentry instead of NXroot
```

which felt odd to me (`/` is often `NXroot` and not `NXentry`).

A quick check in the code showed me that the NXclasses in the error message were indeed swapped. This PR fixes this.


vasole:
@woutdenolf That piece of code is your baby and I do not use it much, please take care of the review.

vasole:
@loichuder 

If the message says / is an instance of NXentry and not of NXroot, that means the file is not properly written.

If the file was written by PyMca, it is a bug in PyMca (or the user). What program wrote the file?

woutdenolf (approved):
LGTM

Issue #963, opened 2023-02-16
[Packaging] Add cx_Freeze script


vasole:
It seems cx_Freeze is still maintained.

vasole:
closed by #975

PR #975, opened 2023-03-22
[Packaging] Recover cx_Freeze packaging

vasole:
It solves autocompletion issues with the interactive console.


vasole:
closes #963

PR #979, opened 2023-04-13
[IO] HDF5 Stack. Solve problem of selecting particular scan number at the ESRF

vasole:
If the selection matches an entry name, assume one is dealing with entry names in the selection.

### DIFF
--- a/PyMca5/PyMcaIO/HDF5Stack1D.py
+++ b/PyMca5/PyMcaIO/HDF5Stack1D.py
@@ -2,7 +2,7 @@
 #
 # The PyMca X-Ray Fluorescence Toolkit
 #
-# Copyright (c) 2004-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2023 European Synchrotron Radiation Facility
 #
 # This file is part of the PyMca X-ray Fluorescence Toolkit developed at
 # the ESRF by the Software group.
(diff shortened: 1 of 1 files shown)


PyMca5/PyMcaIO/HDF5Stack1D.py
```diff
@@ -214,8 +214,14 @@ def loadFileList(self, filelist, selection, scanlist=None):
...
-                JUST_KEYS = True
+                # the ESRF uses "1.1" notation for the scans so this is ambiguous because
+                # one does not know if the selection is based on scan name or scan order
+                if scanlist[0] in entryNames:
+                    # assumed the selection is based on scan name
+                    JUST_KEYS = False
+                else:
+                    number, order = [int(x) for x in scanlist[0].split(".")]
+                    JUST_KEYS = True
             except:
```
woutdenolf, line 225: Bare exceptions like this will capture all `BaseExceptions` like `KeyboardInterrupt` and `SystemExit`.

vasole: In 17 years of code maintenance I have seen exceptions changing type and therefore not being caught.

In situations where multiple exceptions can happen, I would prefer to explicitly catch `KeyboardInterrupt` and `SystemExit` to re-raise them rather than missing any possible exception.

You see that philosophy in PyMca when importing optional dependencies. Instead of checking for ImportError, I prefer to catch any exception to avoid situations like https://github.com/silx-kit/silx/issues/3731

vasole: Are there any other base exceptions than `KeyboardInterrupt` and `SystemExit` that you would like to get propagated?

woutdenolf: A bare except

```
try:
    ...
except:
    ...
```

is equivalent to

```
try:
    ...
except BaseException:
    ...
```

If you want to capture all non-lethal exceptions you can do

```
try:
    ...
except Exception:
    ...
```

This ensures you don't capture exceptions that should never be _absorbed_ (i.e. if you capture them you should ensure a re-raise).

woutdenolf: For this reason any linter like `flake8` warns against bare excepts.

vasole: OK. I will change `except:` to `except Exception:`

vasole: #980 addresses it.

PR #1064, opened 2024-03-03
Proof of concept implementation: Tiled as DataSource in PyMca

linupi:
This is a proof of concept implementation using [Tiled](https://blueskyproject.io/tiled/) as DataSource in PyMca. It is implemented in a similar way as the Redis access for [BLISS](https://bliss.gitlab-pages.esrf.fr/bliss/master/index.html). 

This branch is not intended to be used in production (and not to be merged). It's only purpose is to explore the feasibility of linking  `PyMca` and `Tiled`.

tiled_in_pymca2
Similarly as for `BLISS` and `SPEC`, `Tiled` can be accessed through the button in the side navigation.

tiled_in_pymca
Scans that are visible here are directly served via `Tiled`.


vasole:
I do not mind to ship PyMca with specfile-like support for Bluesky as it is done for bliss. It will be limited to what can be wrapped as SPEC scans, but if users find it useful...

By other hand I have just merged the corrected h5py-like support. It was needed to read formats supported by silx but not by PyMca.  To access tiled, it would still need the tiled branch of t20100 merged into silx.

PR #1082, opened 2024-06-26
merge dev-ab into master

alexbaekey:


vasole:
Hello,

Please let me know if you want me to do something. This looks to me as tests from your side.

Once you are satisfied with the support of your file format, I can merge that part just taking care that it does break anything (for instance your test for being a TAS file should go at the very end because just testing for a `#`  is a bit light).

PR #1087, opened 2024-08-14
Src layout

woutdenolf:
Closes #1086

bootstrap.py is not needed anymore so I guess we can remove it as well.

PR #1088, opened 2024-08-14
Add motor info to QNexuswidget

mspito:
Needs https://github.com/vasole/pymca/pull/1087 first

Screenshot from 2024-08-14 16-15-52

### DIFF
--- a/src/PyMca5/PyMcaCore/NexusTools.py
+++ b/src/PyMca5/PyMcaCore/NexusTools.py
@@ -510,6 +510,28 @@ def getPositionersGroup(h5file, path):
                 positioners = group
     return positioners
 
+def getStartingPositionersGroup(h5file, path):
+    """
+    Retrieve the start positioners group associated to a path
+    retrieving them from the same entry.
+
+    It assumes they are either in:
+
+    - NXentry/NXinstrument/positioners_start or
+
+    """
+    entry_path = getEntryName(path, h5file=h5file)
+    instrument = getNXClassGroups(h5file, entry_path, ["NXinstrument", b"NXinstrument"], single=True)
+    positioners = None
+    if len(instrument):
+        instrument = instrument[0]
+        for key in instrument.keys():
+            if key in ["positioners_start", b"positioners_start"]:
+                positioners = instrument[key]
+                if not isGroup(positioners):
+                    positioners = None
+    return positioners
+
 def getMeasurementGroup(h5file, path):
     """
     Retrieve the measurement group associated to a path
--- a/src/PyMca5/PyMcaCore/NexusTools.py
+++ b/src/PyMca5/PyMcaCore/NexusTools.py
@@ -515,7 +515,11 @@ def getStartingPositionersGroup(h5file, path):
     Retrieve the start positioners group associated to a path
     retrieving them from the same entry.
 
-    This method assumes the positioner group is NXentry/NXinstrument/positioners_start.
+    It assumes they are either in:
+
+    - NXentry/NXinstrument/positioners_start or 
+    - NXentry/NXinstrument/positioners or
+    - NXentry/measurement/pre_scan_snapshot
 
     """
     entry_path = getEntryName(path, h5file=h5file)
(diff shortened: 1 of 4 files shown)


vasole:
Please rebase your work on current master.

vasole:
Some comments:

PyMca is used at other facilities than the ESRF and former ESRF files did not look like current ones.
I find  `nxpositioners = nxentry.get("instrument/positioners_start", None)` very ESRF-centric

I would aim at using the function getPositionersGroup(h5file, path) in PyMca5.PyMcaCore.NexusTools.py that would make the use of the new feature not restricted to ESRF. That would imply that if one positioner is an array of positions, the first position is used as starting point. A getStartingPositionerValues in PyMca5.PyMcaCore.NexusTools.py seems to me more appropriate than get_motor_positions.

mspito:
I tried to address your suggestions. Is there anything else to be done?

src/PyMca5/PyMcaCore/NexusTools.py
```diff
@@ -510,6 +510,28 @@ def getPositionersGroup(h5file, path):
...
     return positioners
 
+def getStartingPositionersGroup(h5file, path):
+    """
+    Retrieve the start positioners group associated to a path
+    retrieving them from the same entry.
+
+    It assumes they are either in:
+
+    - NXentry/NXinstrument/positioners_start or
```
woutdenolf, line 520:
```suggestion
    This method assumes the positioner group is NXentry/NXinstrument/positioners_start.
```

woutdenolf (changes requested):

vasole:
Prior to it there was a group inside NXinstrument named `positioners` and no `positioners_start`

vasole:
> What makes you think `positioners_start` is "new"? As far as I can remember it has always been there.

If it serves as reminder, `silx convert` is still generating the `NXinstrument/positioners` group.

woutdenolf:
For reference: `NXinstrument/positioners` provides motor positions in the scope of entire scan. This means static motors are scalars and moving motors are arrays. `NXinstrument/positioners_start` and `NXinstrument/positioners_end` are snapshots of motor positions at the start/end of the scan and are therefore scalars.

vasole:
> For reference: `NXinstrument/positioners` provides motor positions in the scope of entire scan. This means static motors are scalars and moving motors are arrays. `NXinstrument/positioners_start` and `NXinstrument/positioners_end` are snapshots of motor positions at the start/end of the scan and are therefore scalars.

`positioners` is always there while start and end are not always there . Just convert a SPEC file with silx convert to verify it. Therefore, I want to check first for positioners_start, and if not present, to use positioners, and if not present to try the Sardana equivalents. It is a 10 minutes work. Simply I do not consider to be my job to do it.

woutdenolf:
So what needs to be done is

- remove `getStartingPositionersGroup` and use `getPositionersGroup` in `get_motor_positions` instead, assuming the datasets can be 0D or 1D.
- modify `getPositionersGroup` to look for `positioners_start` before looking for `positioners`.

Is that correct?

vasole:
One cannot escape to having two different functions (or a function with keywords).

You need to keep into account that to match the SPEC behaviour, you want the motor values associated to the positions when the command was issued. If the information is not available, then one uses the group provided by getPositionersGroup

What you propose should work for this issue, but when I look for the positioners group in PyMca, I actually want the group containing the scanned ones because I want the positions during the scan and not when the command was issued.

So, I would just extend the getStartingPositionersGroup to handle the case positioners_start is not there and it gets the getPositionersGroup instead. The handling of 1D data instead of 0D would indeed be necessary. If you want to centralize the dimensions handling as getStartingPositionersAndValues, I am pretty sure it can avoid duplicating code somewhere else.

vasole:
> We added the function _getStartingPositionerValues_ which calls the function _getStartingPositionersGroup_ which falls back to the function _getPositionersGroup_. Is that what you had in mind to be more generic ?

Yes, it is. Please, just make the new functionality available just at NXentry level.

woutdenolf:
We thought it would be easier if you could click anywhere in the scan and see the list of motors.

mspito:
Now we can access to the motor positions only at NXentry level. Is that what you expected ?

Issue #1093, opened 2024-11-19
[DOC] Incomplete documentation for Windows installation


mretegan:
In the current documentation, it is stated that no additional packages are required when installing on Windows. However, this is incorrect as without the Microsoft Redistributable C++ package the application is missing required DLLs and crashes at start.

PR #1094, opened 2024-11-19
Update documentaion for Windows stand-alone app

mretegan:
Fix for #1093

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -10,7 +10,9 @@ make use of different features of *PyMca* in your own Python programs.
 Stand-alone Executable
 ----------------------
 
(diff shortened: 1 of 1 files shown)


vasole:
That only happens in freshly installed machines where no other C++ application has been installed.

It is not needed to install the runtime systematically. Please, can you specify something like "in case of crash on startup, please make sure that..."

mretegan:
Done. Feel free to change as you wish. The way it was before was causing issues with the new beamline computers.

vasole:
> Done. Feel free to change as you wish. The way it was before was causing issues with the new beamline computers.

I guess the same issue would apply to the `silx`  binary.

mretegan:
Probably, but silx is not a de facto application on ESRF beamline computers, and I assume the majority of people are using PyMca instead of silx for looking at data files.

PR #1110, opened 2025-04-01
use one single function to sort HDF5 items

woutdenolf:
- Move and refactor the three copies of `h5py_sorting` to a single `PyMca5.PyMcaIO.HDF5Utils.sort_h5items` function
- Improve sorting by name by splitting on all non-numeric characters (if any)
- When sorting by title, start_time, etc. has duplicate values, fall back to sorting by name


vasole:
Is it ready from your side?

PR #1116, opened 2025-04-14
Undefined variable dynamically loading a ROI Imaging saved stack.

vasole:
XDatasetList was undefined in case of dynamically loading a stack not fitting into memory and saved by PyMca.


vasole:
closes #1117

woutdenolf (approved):

PR #1122, opened 2025-05-13
[IO] Improve image stack handling

vasole:
The code as it was written was only dealing with a single dataset of images.

This PR makes sure if multiple scans and datasets are selected, they are properly summed.

It falls back to the case the data were not fitting into memory that was lost when avoiding a loop.


woutdenolf (approved):
Oef tricky stuff

PR #1124, opened 2025-05-17
OutputBuffer datasets cannot be the same

### LINKED ISSUE #1123: OutputBuffer issue when nosave=True
Reproduce bug

```python
import numpy
from PyMca5.PyMcaIO.OutputBuffer import OutputBuffer

data = numpy.zeros(10)

output_buffer = OutputBuffer(nosave=True)
output_buffer.allocateMemory("label1", data=data)
output_buffer.allocateMemory("label2", data=data)
output_buffer["label1"][0] = 1
assert output_buffer["label2"][0] == 0
```

This happens whenever HDF5 is not selected as output format. For example by `nosave=True`.

woutdenolf:
Closes #1123 

FYI the `OutputBuffer` is used by PyMca to send fit results to. The buffer takes care of allocating space in memory or HDF5 and saving the fit results in HDF5, EDF, TIF, DAT or CSV.

This bug concerns a pure in-memory space allocation: it does not ensure that each dataset is independent. So if I change the content of one dataset it does not affect another.

### DIFF
--- a/src/PyMca5/PyMcaIO/OutputBuffer.py
+++ b/src/PyMca5/PyMcaIO/OutputBuffer.py
@@ -356,6 +356,8 @@ class OutputBuffer(MutableMapping):
         """
         if data is not None:
             buffer = numpy.asarray(data, dtype=dtype)
+            if id(buffer) in set(map(id, self._buffers.values())):
+                buffer = buffer.copy()
             if fill_value is not None:
                 buffer[:] = fill_value
         elif fill_value is None:


sergey-yaroslavtsev:
```
buffer = numpy.asarray(data, dtype=dtype)
if id(buffer) in set(map(id, self._buffers.values())):
```
`buffer` is a new variable, thus, it should appear with a new id. This id could not be in previously defined `_bufferes` except of overflow, or did I missed smthg?

vasole:
> `buffer` is a new variable, thus, it should appear with a new id. 

With `numpy.asrray`, the copy is made only if needed.

https://numpy.org/doc/stable/reference/generated/numpy.asarray.html

sergey-yaroslavtsev (approved):

PR #1131, opened 2025-06-21
fast fitting: simplify loop logic

woutdenolf:


vasole:
If you have repeated the fit for all the elements twice, you surely are in an endless loop. Therefore a limit based on the number of fitted peak areas.

vasole:
> We have been using `FastXRFLinearFit` with a patched `OutputBuffer` for a while in ewoks to have full control over the output but we had to disable refitting since it's a one-way street. In addition this code was insanely complex. So I decided to bite the bullet and use `ClassMcaTheory` with `ConcentrationsTool` directly.

The original/legacy implementation without OutputBuffer was simpler and on those lines.

PR #1132, opened 2025-06-21
fix shadowed variable

woutdenolf:


src/PyMca5/PyMcaPhysics/xrf/FastXRFLinearFit.py
```diff
@@ -810,17 +810,16 @@ def _fitLstSqNegative(self, data=None, freeNames=None, nFreeBkg=None,
...
 
             # Combine with masks of all other peak areas
             # (unless none of them has negative pixels)
             # This is done to prevent endless loops:
             # if two or more parameters have common negative pixels
             # and one of them remains negative when forcing other one to zero
-            for iFree, (nNeg, iFree, negMask) in enumerate(negList):
-                if iFree not in badParameters and nNeg:
+            for nNeg, iFree, negMask in negList[1:]:
+                if nNeg:
```
woutdenolf, line 822: The first iFree was shadowed.

src/PyMca5/PyMcaPhysics/xrf/FastXRFLinearFit.py
```diff
@@ -810,17 +810,16 @@ def _fitLstSqNegative(self, data=None, freeNames=None, nFreeBkg=None,
             # Bad pixels: use peak area with the most negative values
             negList.sort()
             negList.reverse()
```
woutdenolf, line 812: `negList.sort(reverse=True)` seems to exist but the numpy docs don't mention it. So I didn't touch it.

woutdenolf: Oef, we are talking about a list. So we can use it.

vasole: You can reverse a list without sorting it first, but the output is not necessarily the same.

image

src/PyMca5/PyMcaPhysics/xrf/FastXRFLinearFit.py
vasole: It is quite long ago, but I think you are missing the point of pixels already included in the list to be re-fitted. One pixel can have more than one element with negative areas.

vasole: OK. Got your point.

vasole: Then I have to figure out is why the enumeration was necessary. It makes no sense if it is not to be used.

PR #1136, opened 2025-07-18
skip too long datasets to prevent crush

sergey-yaroslavtsev:
When click on dataset in the group it adds all datasets to the table. If one dataset is too long it does not fit QSpinBox which leads to fact that any dataset from the group could not be displayed.
Rare and funny bug. 
I suggest just to skip such long datasets as one will not "spin" over 2 billion of layers.

Screenshot 2025-07-18 142213


woutdenolf:
Can you show me a screenshot of the "spinbox"? I have never seen it. Which window does it appear in?

vasole:
Something like this:

image

sergey-yaroslavtsev:
> Something like this:
> 
> image

would it more shorter to:
```
        if maximum and maximum < 2**31-1:
            self._index.setMaximum(min(maximum,2**31-1)  - 1)
```

woutdenolf (approved):

PR #1138, opened 2025-07-30
Rgbcorrelator selector fix

sergey-yaroslavtsev:
Few small fixes for file selector of RGBCorrelator.
1) on first open add all available types it was:

image

2) if all images were removed using red-cross button image size was still in memory not allowing to select an image with another size

image

3) Add possibility to have multiple selection inside h5fille navigator. It is useful to select few out of many in one group, or to select few groups at a time.


sergey-yaroslavtsev:
sorry mess up took corrections from previous branch instead of master.

PR #1199, opened 2026-02-24
Add flexibility to selection table

sergey-yaroslavtsev:
To close two jira requests:
https://jira.esrf.fr/browse/DPDEV-387
https://jira.esrf.fr/browse/DPDEV-346
Add 2nd Y-axis selection (map to right)
Add multi monitor selection - signal/multi selection optional. 

it is a bit of copy-paste the original "Signal" 
From brief test it works quite nice. But maybe I am missing some crucial point? in code it was explicitly written "only one monitor".


sergey-yaroslavtsev:
Rebased.
Now in `HDF5CounterTable` new features are disabled by default.  They are explicitly enabled only in `PyMcaNexusWidget`.
`Signal R` to appear is now require user interaction -> r-click on `Signal` -> "two Y-axis".
`Signal R` accepts only 1D data explicitly - there is no point for second Y-axis if user try to select 2+D data.

sergey-yaroslavtsev:
Check if main signal is not 1D then `Signal R` should not be used at all (message pop-up) to avoid conflict.

woutdenolf (changes requested):

woutdenolf (approved):

PR #1209, opened 2026-04-09
Solve the locking issue

sergey-yaroslavtsev:
locking mechanism should be set before "import" not before "create". 
At least in demo_session of Bliss, the PyMca was blocking writer. With this correction it works OK.

Or did i miss the point? Sorry i am still quite confused with multiple accesses.


woutdenolf:
> locking mechanism should be set before "import" not before "create".

Why? Are we opening HDF5 files before import? I doubt it.

woutdenolf (approved):

PR #1211, opened 2026-04-16
Refresh button rethink + auto-refresh for HDF5 files

### LINKED ISSUE #1210: Auto-update of the plot and Refresh button
Some time ago PyMca was able to show updating image (same what flint is doing now). I hope it is not my bad memory + imagination. It would be nice to restore this functionality since some users complain about it. 

Partially related thing is that `refresh` button do not update the current plot. When refresh is clicked  - it collapse the h5 tree (which is annoying) and deselect the selected scans (which again could be frustrating if user work with several). 

Another proposition was to whenever any dataset is clicked the refresh procedure is triggered. I could see why it is ugly but it is quite understandable from user perspective. See DPDEV-393 for example.

sergey-yaroslavtsev:
Closes #1210
I made a single PR because logic of these two things is convoluted. 

big changes:
1) Refresh button now do not collapse the HDF5 tree and do not deselect datasets and replot at the end of refresh.
* by adding methods to save/collect/restore the HDF5 tree from the `self.hdf5Widget` 
2) 'Auto' checkbox added to update data from HDF5 file every second. It do not update the tree (so new datasets will not appear) but reread the data and replot.
* trick was not to call the refresh every time since it is time-consumable (also related to new methods of tree save-restoration), then to avoid call of destructor (to avoid tree collapse) and swap `file` in nodes. Yes it is kind of a hack - but i do not see it to be fragile. 

small things;
1) fix some bugs (they were needed for main features or for their tests):
   * `sys.exc_info()[1].args[0].upper()` could appear to be integer, thus, `upper` raise an error.
   *  `delta = max(delta, abs(ylength - mlength))` the `abs` was missing leading to error instead of desired stripping. (it is exreamly crucial for 'auto-refresh' when plot have Monitor(s).
   * `HDF5Utils.safe_hdf5_group_keys(file_path, data_path=data_path)` could return empty and, thus, avoiding proper handling
2) refresh button now have a hotkey - F5
3) refresh button avoid the queue (if user click it while refresh is still in process)
4) some loggers were switched from `info` to `debug` to avoid spam in terminal during `auto-refresh`

I've checked in the Bliss-demo-session. Did not found issues for now.

I think this PR is highly desired by many people. Thus, if you can think of better approach please let me know. Of course one can introduce BlissData into PyMca, then even SCAN[-1] could be read from Bliss and not from HDF5 file but it will require much more work...


src/PyMca5/PyMcaGui/io/hdf5/HDF5Widget.py
```diff
@@ -406,11 +406,17 @@ def raw_keys(self):
                     return self.file[data_path].keys()
                 else:
                     from PyMca5.PyMcaIO import HDF5Utils
-                    return HDF5Utils.safe_hdf5_group_keys(file_path,
+                    result = HDF5Utils.safe_hdf5_group_keys(file_path,
                                                       data_path=data_path)
+                    # may have failed silently returning empty
+                    # if file locked by a writer - happens on Windows at least
+                    if result:
```
woutdenolf: An empty list or `None`? Because an empty list can be an empty HDF5 group which is valid.

woutdenolf (changes requested):
Would it be possible to have the updating of the file handles and caching of the tree state in `HDF5Widget` itself?

woutdenolf (approved):
Thanks that is better. All good for me.

PR #1213, opened 2026-04-24
fix silx version to 2.2.2 for wheels testing

sergey-yaroslavtsev:
New version of `silx` require to install in addition `qtawesome`. 
I would suggest currently to use stable 2.2.2 version for testing instead of 3.0.0
Later if no issue will be found we can update version and add `qtawesome`.

failed dry-run release:
https://github.com/silx-kit/pymca/actions/runs/24893920880
python 3.9 install 2.2.2 and run without issues.


t20100:
Why not having `extra_requires` targets: for all dependencies (where you would have `silx[full_no_qt]` and for tests?

t20100 (approved):
LGTM

PR #1220, opened 2026-06-11
key priority sorting for HDF5 files

### LINKED ISSUE #1195: Problem with sorting in HDF5.
filtering by double-click on `Description` is not working properly in new version.
5.9.4 vs 5.9.6:

Image

filtering by double-click on `File/Group/Dataset` works the same in both

### LINKED ISSUE #1217: HDF5Utils.sort_h5items change in 5.9.5+
We have some files that usually contain `['/dark_current', '/mono_0', '/pvs']` as the top-level entries.

In `HDF5Utils.sort_h5items`, the `list_to_sort` gets set the following by `_extract_sort_key_from_name`:

```
[('/dark_current', ('dark_current', <HDF5 group "/dark_current" (1 members)>)), 
((0,), ('mono_0', <HDF5 group "/mono_0" (31 members)>)), 
('/pvs', ('pvs', <HDF5 group "/pvs" (1 members)>))]
```

Then we get a `TypeError` when making the `sorted_list` on line 124 because you can't compare a string (`'dark_currrent'`) and a tuple (`(0,)`)

Introduced in #1110 

Thanks!

sergey-yaroslavtsev:
Closes #1217 
Closes #1195
I redo sorting in priority instead of using single key.  
Using `_NUMERIC_CHARS` instead of `_NON_NUMERIC_CHARS ` because then text will be compared with text and numbers with numbers (no matter what are the names).
Before if key is missing for at least one item sort by that key was skipped.
Now it just sorts by every key in `sorting_list` in a priority order. If 'title' is there it is _forced_ to be highest priority.
I did tests on simulated data (with partially missing `start_time` and with different titles) seems OK for me...


woutdenolf (changes requested):

woutdenolf (changes requested):

woutdenolf (approved):
LGTM

PR #1227, opened 2026-06-25
allow showing scans with differnt motors

sergey-yaroslavtsev:
It was request by ID26. 
Scans of same counter but different motors need to be plotted at the same time.
Scans are exactly the same except of motor name.
Actually it was already working but there was annoying pop-up error - but plotting was working for two scans.

This PR explicitly tries to use the same positioner (not by name but by counter number) or the first one - otherwise fails as before.


vasole:
Honestly, I fail to see what you are trying to correct.

Are you aware of the feature allowing multiple plots of alternating "entries" when one of the entries did not feature the scanned motor? It requests a minimum amount of entries selected, because when having just 3 scans, for instance, it is not cumbersome to select those needed. If you replace the missing motor by another one based on the index of the first one, you are going to break more important things.

From my understanding, what you show there it just requires to be in AUTO ADD mode.

vasole:
> Before it was failing on next line `data = phynxFile[path]` producing an error when selecting several plots. The plot was working! but selecting one more was killing previous one...

The feature allowing an incompatible selection to be applied on multiple entries ignoring the conflicting ones requires the same type of entry at start and end. The fact of ending the selection with a compatible or incompatible entry changes the behavior. It is intended behavior implemented to allow multiple selections of x.1 and x.2 scans because the x.2 was not containing the same datasets and it was cumbersome to alternate the selection by hand above a minimum of scans. For instance, if you start the selection with 1.1 and you end with 5.1 it will allow a multiple selection of scans even if incompatible (by ignoring the incompatible ones) while if you end with 5.2 it will complain.

PR #1239, opened 2026-08-04
Validator and refactor of HDF5Wizard

### LINKED ISSUE #1238: Data selection for ROI image
There are several strange behaviors when selecting datasets as `signal` and `axes` for ROI image. For example dataset of `NxM` can be open with `grid` (not `scatter`) when `1d data is last dimension` but could not be when `1d data is first dimension` . There is a strange silent (for user) behavior if dimensions of dataset contain singletons (some times they are dropped) - is probably done for a reason (some strange data structure) but it is not clear when user do selection - and it can be incompatible with some options. There was also a "hidden" feature to open 2D images (when dataset have no channels - or in other words there is only one "imaginary" channel) which was not working `considerAsImages`.

HDF5Wizzard start to have validation check for `scatter` mode so now i am sure it is required in a proper form for all selections. 

To open images in ROI image could be a good thing but it should not be done silently - since we have already options `1D data is first dimension`  `1D data is last dimension` we can simply add `no 1D data (image)`. 

Drop of singletons should be simple `Yes/No` question after selection. (It should happen before all validations ofc)

Already start work on it.

sergey-yaroslavtsev:
Close #1238

positive:
1) many validators were added so the user have a info about what is wrong and even suggestions what was selected wrong. + it sends back to Wizard instead of closing it with error.

2) clear use of "1D data" position. Now there is no _hidden_ (from UI perspective) possibility to open 2D image. To open images as stack "No 1D data" can be selected.
image

3) singletons are dropped or kept automatically if possible; in cases where it is not obvious the user is asked what to do with them.

4) if axis are bigger than dataset - padding is proposed (or cancel to reselect).

other changes:
a) pure singleton is not acceptable - since it makes no sense and would require a special handler. 
b) order of axes matters - it is tipped in selection - it was done because before if the number of channels was equal to the size of dimension - it was unclear what will happen.

dimension validator gives:
```
| shape                             | first  | last   | none   |

| `(1,)`                            | REJECT | REJECT | REJECT |
| `(A,)`                            | REJECT | REJECT | ACCEPT |
| `(1,1)`                           | REJECT | REJECT | REJECT |
| `(1,A)` / `(A,1)`                 | ACCEPT | ACCEPT | ACCEPT |
| `(A,B)`                           | ACCEPT | ACCEPT | ACCEPT |
| `(1,1,1)`                         | REJECT | REJECT | REJECT |
| `(1,1,A)` / `(1,A,1)` / `(A,1,1)` | ACCEPT | ACCEPT | ACCEPT |
| `(1,A,B)` / `(A,1,B)` / `(A,B,1)` | ACCEPT | ACCEPT | ACCEPT |
| `(A,B,C)`                         | ACCEPT | ACCEPT | REJECT |
```
then there is next validator based on selected axes. The full table of every possible outcome will be quite big.

NOTES:
Some things could be not obvious on first place but there is a logic - for example:
if dataset have singletons and "No 1D data" is selected. Singletons are dropped automatically because having them would only restrict dimensions - while without them user can reshape - and default option is the one as if they do exist (it is for free).

I used some AI generated tests to verify obvious things but i will test more things manually - that is why it is [WIP].

Feel free to suggest different logic. I am also fine if you would sat that this PR is bad idea - but then please provide an idea(s) to fix current state.


woutdenolf (changes requested):
Looks good overall. Some small suggestions.

woutdenolf (approved):

PR #1242, opened 2026-08-21
Protect the scatter mode to have Silx

sergey-yaroslavtsev:
New `scatter mode` implementation in ROI tool require silx - since silx is not mandatory the protection is required.


vasole (approved):
It works, thank you.

BTW, line 1349 seems to have a typo in the comment.

PR #1243, opened 2026-08-21
test import erros in specific cases

sergey-yaroslavtsev:
TestAll and CLITest ignore import errors because there are OS specific libraries which could create problems. 
This test try import PyMca modules bypassing this rule but only for particular modules which are utilized via scripts. 

1) Ideally this PR should now **fail** all the tests - because it require #1242 to be merged

2) After #1242 will be merged - rebase - this PR should become green.

3) To be sure before merging i would run `dry-run` to be sure no OS related issues appear because of this test.

Do not merge it before these three steps are done.


sergey-yaroslavtsev:
some configuration of ubuntu have problem not only with `silx` (what was expected) but also with `ImportError: libEGL.so.1` 
difference between Qt6 and Qt5 configurations

sergey-yaroslavtsev:
seems that Qt6 needs more Linux graphic libraries...
```
              libxcb-util1 \
              libxcb-cursor0 \
              libglu1-mesa \
              libgl1 \
              libegl1 \
              libx11-xcb1 \
              libxi6 \
              libxrender1 \
              libsm6
```
actually it probably means that it was kind of a bug in tests... because tests could have been skipped in CI because UI was not working based on import error.

sergey-yaroslavtsev:
now all CI tests fails only because of `silx` as expected

sergey-yaroslavtsev:
All CI tests pass after rebase with #1242

sergey-yaroslavtsev:
`dry-run` started at https://github.com/silx-kit/pymca/actions/runs/32488173152

vasole:
I guess that in analogy with the widget instantiation tests, the presence of a Qt binding should be considered prior to try those specific test imports that require Qt.

See:

https://github.com/silx-kit/pymca/blob/master/src/PyMca5/tests/WidgetsInstantiationTest.py

sergey-yaroslavtsev:
> I guess that in analogy with the widget instantiation tests, the presence of a Qt binding should be considered prior to try those specific test imports that require Qt.

Like this would be good?

Every check which I wanted - pass.

vasole (approved):
Thank you. Fine with me.

PR #1244, opened 2026-08-21
fixing plot title for NNMA

sergey-yaroslavtsev:
The reported bug seems to be fixed


vasole (approved):
Thank you. It works now.

PR #1249, opened 2026-09-02
try to read HKL from motors if missing

sergey-yaroslavtsev:
Read HKL from motors for the SCAN plot the same way as the `motor info` plugin do.
Copy-paste of reading procedure not to import the Plugin + searching for H K L or h k l.

image


vasole:
The correct way is to add the information into the `info` part returned by NexusDataSource.getDataObject using the NexusTools.py module in analogy with what is done in SpecfileDataSource.

If one does not implement the information at the source, one ends up having to patch code at multiple places.

I would aim at extracting the HKL information from NexusTools.getStartingPositionerValues because what is shown there is the HKL position at the beginning of the scan.

The place in NexusDataSource.getDataObject would be around the position shown by the lines below:

image

sergey-yaroslavtsev:
Seems that  `getStartingPositionersGroup` is enough, `getStartingPositionerValues` read more than needed. 

One point to be mentioned:
in case if HKL is changing should we show starting values? I mean user can misunderstood that these are values for the whole SCAN when they are only starting values... No idea what is expected by the users.

sergey-yaroslavtsev:
`hkl.append(float(dataset[()]))` 
could be changed to 
`hkl.append(float(dataset[(0,) * dataset.ndim]))`
I think it would be worse... but not sure...

vasole:
> One point to be mentioned:
> in case if HKL is changing should we show starting values? I mean user can misunderstood that these are values for the whole SCAN when they are only starting values... No idea what is expected by the users.

It was the same situation in the past with SPEC hklscans. Nothing new.

Only the HKL position when the command was issued (if available). If more than one HKL position available, the first HKL position is to be shown at that particular place.

It is what is expected by the users at that place because it is what they always had.

sergey-yaroslavtsev:
It should work as expected. We are reading starting positioners and even if array appears there we are taking the first element. 

I test it on the data from BM28 - seems to work. Switching from one plot to another also change the HKL. 

Something else should be changed?

vasole:
Please, can you `silx convert` one of the beamline SPEC files to HDF5 and see if the information is correctly displayed?

I am not 100 % sure about `silx convert` writing the HKL information within the `positioners` or the `sample` group.

sergey-yaroslavtsev:
> beamline SPEC files

I do not have such... So I made few things to test. If it is not OK let me know I will ask around to find the _real_ file.

If H K L are declared as pseudo motors - they could be found as proper positioners. And everything works.
If H K L are declared as `#Q` conversion ignores it - they appears only in `instrument/positioners/specfile/scan_headers` as a string "#Q 0 0 1".

Works:
```
#S 1  scan  0 0  0 0  1 3  10 1
#D Thu Feb 11 09:55:20 2016
#T 1  (Seconds)
#O1 H  K  L
#P1 0 0 1
#L A  B  C  Epoch  Detector  Monitor
0 0 1.0000 100 1000 10000
0 0 1.2000 101 1030 10000
```

Does not:
```
#S 1  scan  0 0  0 0  1 3  10 1
#D Thu Feb 11 09:55:20 2016
#T 1  (Seconds)
#Q 0 0 1
#L A  B  C  Epoch  Detector  Monitor
0 0 1.0000 100 1000 10000
0 0 1.2000 101 1030 10000
```

vasole:
> I do not have such... So I made few things to test. If it is not OK let me know I will ask around to find the _real_ file

The best is to ask the person that reported the issue (I do not want to write the name here) to provide a SPEC file with meaningful HKL values. The files I have found contain only zeros.

vasole:
OK. I have found a very old SPEC file.

`silx convert` does not succeed on recovering the HKL as positioners (only as counters) therefore you can ignore that test.

vasole:
Please, can you add the scan entry and title (only if present) in the scan field to the displayed information in analogy with SPEC?

image

vasole:
If you prefer to be in a separate PR...

sergey-yaroslavtsev:
`title` is the command and it was already in the `info`, but to mimic SPEC we need also the scan number - in HDF5 it appears to be the `entry` itself -> passed to `info` then read and split in the widget.
I decide to remove initial "/" just for beauty.

image

vasole (approved):
Thank you!

